Issue #4, opened 2019-02-28
Building the basic example fails


ehsanmok:
Hi

First of all, thank you for the binding initiations. Rust needs to have it eventually :)

Running `cargo run --example basics` fails using `stable(1.0) cpu libtorch`

```
--- stderr
CMake Error at /home/ehsan/LibTorch/libtorch/share/cmake/Caffe2/public/utils.cmake:17 (add_dependencies):
  add_dependencies Cannot add target-level dependencies to INTERFACE library
  target "caffe2_library".

Call Stack (most recent call first):
  /home/ehsan/LibTorch/libtorch/share/cmake/Caffe2/Caffe2Config.cmake:121 (caffe2_interface_library)
  /home/ehsan/LibTorch/libtorch/share/cmake/Torch/TorchConfig.cmake:39 (find_package)
  CMakeLists.txt:4 (find_package)
```

I think, we should consider adding a CI.

LaurentMazare:
Hi,
Thanks for the feedback, yes CI would certainly be a good thing to add.
Re your compilation error, I haven't encountered it despite trying the pre-built cpu version 1.0.0 and 1.0.1 (as per build-version).
Is it on linux or on another platform ? I only tried linux on my side.

ehsanmok:
Yes, it was on Ubuntu 14.04 with cmake 3.2

LaurentMazare:
Interesting, googling a bit the closest I found to your error is this [issue](https://github.com/facebookarchive/caffe2/issues/2281). Not sure whether it's related but if it's the case updating to cmake 3.5.1 may fix the issue.

Alternatively you can try using the `cc` crate rather than `cmake` to compile. This can be done by replacing `build.rs` with the following. If this works for you we can either switching to `cc` or having an environment variable to decide between `cc` and `cmake`.

```rust
use std::env;
use std::path::PathBuf;
extern crate cc;

fn main() {
    let libtorch = env::var("LIBTORCH").expect("LIBTORCH not defined");
    let libtorch = PathBuf::from(libtorch);
    let libtorch_lib = libtorch.join("lib").into_os_string().into_string().unwrap();
    println!("cargo:rustc-link-search=native={}", libtorch_lib);
    println!("cargo:rustc-link-lib=c10");
    println!("cargo:rustc-link-lib=caffe2");
    println!("cargo:rustc-link-lib=torch");
    cc::Build::new()
        .cpp(true)
        .include(libtorch.join("include"))
[...]

PR #8, opened 2019-03-03
Improve the build process

ehsanmok:
This PR contains

1. More idiomatic `build.rs`
2. cmake minimum version fix
3. `.gitignore` cleanup

Tested locally w/o `LIBTORCH` was set and w/o `LIBTORCH_USE_CMAKE` was set.

Note that `LIBTORCH` now should be the root repo containing `libtorch`.

[#4](https://github.com/LaurentMazare/tch-rs/issues/4)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
-/target
-torch-sys/target
+target/
 _build/
 data/
 gen/.merlin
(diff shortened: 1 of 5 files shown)


LaurentMazare:
Thanks for the PR!
I restored the old `LIBTORCH` behavior to be the actual `libtorch` directory as I have some use case where this directory is actually named `libtorch-1.0.1` and it's used by other people so renaming it would be a bit difficult.
Also I like that we don't rely on an external command to download the library anymore but I'm a bit sad about the amount of dependencies that it drags. It's probably fine for now but we may reconsider this in the future.

PR #10, opened 2019-03-03
Normalize format with rustfmt.

ehsanmok:
It's a very common practice to dump `rustfmt.toml` from `rustfmt` to normalize the format in the crate for all the contributors. This needs to be generated once (dumped from `rustfmt 1.0.1-stable (be13559 2018-12-10)`) and is applied to the whole package.


LaurentMazare:
Thanks for the PR! Actually I already used rustfmt but it wasn't installed on my laptop so my changes from the last few days were not formatted properly.
Two points make this a bit tricky:
- I don't think the generated code should be formatted in this PR, I'll have a look at doing this at the code generation level.
- Some comments I've put don't seem to interact well with rustfmt, I'll go over them manually soon.

In order to avoid conflicts, would you make a PR with only rustfmt.toml ? Hopefully the other files will get cleaned pretty soon.

ehsanmok:
Since generated ones are part of this crate, I think they should be formated. I'll let you to decide on that :)

I'll change it by the end of today.

ehsanmok:
After this, the lint should be add to the CI and invoke before any build. It might be possible to configure not to touch the generated ones if you really don't want to normalize their formats (which I don't think is a good idea, since they've already been in the crate). If you don't include them and somehow let them to be generated during the build, then it should be fine.

LaurentMazare:
Thanks again for the PR, I think you're right and the generated files should be formatted but I would like this to be part of the code generation process rather than being done manually in order to avoid large diffs - I will have a go at this soon.

PR #16, opened 2019-03-06
Add torch device interface and clean up.

ehsanmok:
The current device implementation doesn't do proper error handling, since `unsafe_torch!` were used everywhere instead of `unsafe_torch_err!` so for example, `cuda_is_available` or `Device::cuda_if_available()` shouldn't just panic due to any reason in the client code. Also `device_wrapped.rs` doesn't provide any safety and is redundant. 

This PR fixes these issues.


ehsanmok:
@LaurentMazare ping.

LaurentMazare:
Actually I don't think these functions should ever fail on the C++ side so i'm not sure we should make them fallible on the rust side neither. I recently tried to add some fallible versions for the functions that I believe could fail.

ehsanmok:
@LaurentMazare I also checked the ATen related parts and I didn't find any exception handling, so they seem be unwrapped safely. I changed those back to `unsafe_torch!`. Please check.

PR #18, opened 2019-03-06
Add optim trait to make optimizers impl configurable.

ehsanmok:
This PR add the `Optim<Config>` trait so that implementations of different optimizers adhere to a common extensible interface.

@LaurentMazare Please check.


LaurentMazare:
Closing this now as the optimizer have been refactored a bit over the last couple weeks, hopefully with similar ideas. Using `derive_builder` seems nice but I prefer limiting the dependencies for now.

PR #23, opened 2019-03-11
Add support for building with CUDA

mwilliammyers:
On Linux, use `TORCH_CUDA_VERSION` to specify the version of CUDA to build libtorch with.


LaurentMazare:
Thanks!

Issue #30, opened 2019-04-30
element_size_inbytes is not present anymore


infinite-Joy9l:
While trying to build this from a manually downloaded torch library, I got the below error.

```
cargo:warning=In file included from libtch/torch_api.cpp:6:0:
cargo:warning=libtch/torch_api.cpp: In function ‘at::Tensor* at_tensor_of_data(void*, int64_t*, int, int, int)’:
cargo:warning=libtch/torch_api.cpp:48:48: error: ‘class at::DeprecatedTypeProperties’ has no member named ‘elementSizeInBytes’
cargo:warning=     if (element_size_in_bytes != tensor.type().elementSizeInBytes())
cargo:warning=                                                ^
```

Now I can see that this method element_size_in_bytes has been removed recently. https://github.com/pytorch/pytorch/pull/17785, which might be the reason for this failure in the build process

infinite-Joy9l:
I have made the changes in this PR https://github.com/LaurentMazare/tch-rs/pull/31, The above error is gone replaced with a new error of more changes that hint at more api changes. Looking into that.

LaurentMazare:
Thanks I'll have a look at the PR in the coming days.
Please note that we will want to preserve compatibility with the current stable version of PyTorch (1.0.1) so I would suggest also testing your changes with this version. When the stable version of PyTorch gets updated we could consider more breaking changes to accommodate with the new version and not support the old stable version anymore.

infinite-Joy9l:
I am not a C++ expert and so nothing comes to my mind in terms of maintaining compatibility between different versions of libtorch. There seems to be too many api changes between the two libtorch versions.

LaurentMazare:
Actually PyTorch v1.1.0 ended up being officially released yesterday so I just pushed some changes to support this version. Hopefully this should fix this issue.

LaurentMazare:
Closing as the issue should be solved, feel free to reopen if you have any further issues.

PR #31, opened 2019-04-30
Remove support for Type::elementSizeInBytes as no longer available in…

infinite-Joy:
… the libtorch api.

Basically similar changes as can be seen in this PR.


LaurentMazare:
The CI failure is caused by `element_size` not existing in the current stable version of PyTorch.
Maybe it would be possible to have some `#if` pragma to check the version but I haven't been able to find a C pre-processor variable for the libtorch version.
There is a `PACKAGE_VERSION` in some `TorchConfigVersion.cmake` but this cannot easily be used as we don't use cmake for building.
Any thoughts on how to get your change compatible with the 1.0.0 and 1.0.1 version of libtorch ?

LaurentMazare:
The underlying issue #30 should now be fixed.

PR #34, opened 2019-05-06
Make Tensor must_use

Pratyush:
Many methods of `Tensor` return a new `Tensor`; at the moment it's easy to ignore these new `Tensor`s. By making `Tensor` `must_use`, users will get a warning whenever a `Tensor` is left unused.


LaurentMazare:
Nice, thanks for the PR!

Issue #48, opened 2019-05-31
Variable re-using on VarStore


jerry73204:
It's mostly about my thoughts on `VarStore`.

First it's about the way current impl handles duplicated names. As you see this line, it puts a `HashMap` size suffix when collision. However, it leads to inconsistent namings depending the order you add the tensors.

Suppose we want to add three tensors named _foo_, and other three named _bar_.
If we start from _foo_ tensors and then _bar_, it results in
```
foo, foo__1, foo__2, boo, boo__4, boo__5
```
If we add _bar_ tensors first instead, we have 
```
bar, bar__1, bar__2, foo, foo__4, foo__5
```
I'm not sure it's expected or not. It's fine in most cases. I observed this when I wrote tensor logging. Imagine you're developing a model and change the order of model building functions. It could break the log reader. We can take TensorFlow's naming as reference. At least, we can make it order independent.

Another scenario is variable reusing. The current impl always append a new tensor whenever you call methods on Path. Also, I see no way to obtain saved tensors from `VarStore`. That's inconvenient if we want to build two models on shared variables. We may provide an additional "try_reuse" parameter for this.

LaurentMazare:
Thanks for the feedback.
The naming behaviour is indeed on purpose. As you noted it would not work if you change the order in which tensors are created, however even when using an index per base name the same issue would arise if you change the order in which tensors with the same name are created. The idea there is that you should name tensor uniquely for most use cases (except very simple snippets hence this behavior).
Reusing the same tensors would also be problematic if queried with different sizes or initialization or even with the same random initialization as it's unclear whether you should reinitialize or not.

Re variable reuse, the idea there is that you should reuse things explicitly - rather than relying on some context magic as would be the case with tensorflow. If you want to use a variable twice, you should just create it once with for example `let var = path.zeros(...)` and then pass the `var` tensor to all the places where it is going to be used.
I just added a `get` function so that you can get tensors back based on a path and a name, but I think the clean way to do things is more to pass tensors as described previously.

jerry73204:
I agree reusing should be explicitly specified by users. I think passing variables around would lose the convenience of name space hierarchy. Suppose you build two models partially share some params. While you might keep track of bunch of shared variables, `VarStore` would be a good delegation for this case.

I pushed my entry API `nn::Path` and you can take a look. It's aimed for those who want to check name existence beforehand. I'm not sure if this change is crucial for most users, and we can merge after we have plenty discussions.

[...]

PR #49, opened 2019-06-03
Entry API for VarStore

jerry73204:
This PR continues the work on #48 issue. It aims to to provide entry API for `VarStore` for convenience of reusing named variables. It makes sure the explicit semantics of variable creation and reuse.

The example demonstrates the usage of this API. It builds to models with shared model parameters.
```rust
use tch::nn::Path;

impl MyModel {
    fn new(&self, path: &Path) {
        let embedding = path.entry("embedding").or_zeros(&[3, 4, 5]);  // reuse
        // ...
    }
}

fn main() {
    let vs = VarStore::new(Device::Cpu);
    let root = vs.root();
    let orig_model = MyModel::new(&root);
    let mirror_model = MyModel::new(&root);

    // Do separate work
    // ...
}
```

### DIFF
--- a/src/nn/var_store.rs
+++ b/src/nn/var_store.rs
@@ -33,9 +33,11 @@ pub struct Path<'a> {
     var_store: &'a VarStore,
 }
 
+/// An Entry holds an entry corresponding to a given name in Path.
+#[derive(Debug)]
 pub struct Entry<'a> {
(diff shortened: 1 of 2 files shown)


LaurentMazare:
Nice, thanks for the PR and the clean implementation.

PR #53, opened 2019-06-11
Fix the issue that Yolo example cannot compile

jerry73204:
The patch fixes the issue that compiler fails on `vs / x` expressions, where vs is the VarStore and x is a non-&str type.

### DIFF
--- a/examples/yolo/darknet.rs
+++ b/examples/yolo/darknet.rs
@@ -120,7 +120,7 @@ fn conv(vs: nn::Path, index: usize, p: i64, b: &Block) -> failure::Fallible<(i64
     let pad = if pad != 0 { (size - 1) / 2 } else { 0 };
     let (bn, bias) = match b.parameters.get("batch_normalize") {
         Some(p) if p.parse::<i64>()? != 0 => {
-            let vs = &vs / format!("batch_norm_{}", index);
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Thanks for the PR.
Could you tell a bit more what this tries to fix ? The original code compiles fine for me (on rust 1.34.2) but the modified code doesn't (your PR removes a mut that was actually used) and the CI also fails on this change. Is there something I'm missing here ? Maybe you're using nightly or something like this ?

jerry73204:
I compiled Yolo code by providing a standalone `Cargo.toml` with `tch = "^0"` and encountered the error.

 I noticed that crates.io still provides old 0.0.9. After I manually change to git upstream, it works fine. Suppose the crate should be updated to 0.1.0?

jerry73204:
Plus, stable and nightly both compiled or failed. I think it does not matter. The cause is that I downloaded the default 0.0.9.

LaurentMazare:
Got it, I'll try releasing 0.1.0 in the next few days. Closing this PR for now, feel free to open another issue/PR if you run into difficulties (yolo v3 should work out of the box hopefully).

LaurentMazare:
tch-rs 0.1.0 has been released!

Issue #52, opened 2019-06-10
Interop with ndarray


grtlr:
Continue discussion from #51.

Interoperability with the [`ndarray`](https://docs.rs/ndarray/0.12.1/ndarray/) crate would be very nice. As a first step I think we should be able to convert to/from ndarray using `std::convert::Into`. I'm thinking something along these lines:

```rust
use ndarray::{Array, ArrayD, Dimension, IxDyn};

pub fn to_tensor<D>(arr: Array<f64, D>) -> tch::Tensor
where
    D: Dimension,
{
    let tn = tch::Tensor::of_slice(arr.as_slice().unwrap());
    let shape: Vec<i64> = arr.shape().iter().map(|s| *s as i64).collect();
    tn.reshape(&shape)
}

pub fn from_tensor(tensor: &tch::Tensor) -> ArrayD<f64> {
    let v: Vec<f64> = tensor.into();
    let shape: Vec<usize> = tensor.size().iter().map(|s| *s as usize).collect();
    ArrayD::from_shape_vec(IxDyn(&shape), v).unwrap()
}
```

LaurentMazare:
This would indeed be pretty convenient and I think your implementation looks good. One point I'm not sure is whether tensor.into() first calls tensor.contiguous(). If that's not the case it may be worth doing so (and probably good to have a test where we apply from tensor to a 2d tensor that we've sliced on the second dimension).
Would you want to craft a PR for this ? (if you prefer not to, I can also make the change when I get a chance).

grtlr:
I could definitely create a PR for this. It'll be another week or two before I find the time, though.

LaurentMazare:
Cool, let me know if I can be of any help (also if you happen to be at ICML and want to discuss ML in rust, more than happy to do so!).

grtlr:
I will, thanks for the offer! Unfortunately, I’m not at ICML otherwise this would have been very nice!

grtlr:
So I was thinking about how to integrate this in `tch-rs`. Should we include `ndarray` behind a feature flag?

LaurentMazare:
Maybe we can even have it as a normal dependency for now (without any feature flag) ? The number of dependencies for the `ndarray` crate seems quite small so I would hope for it to get compiled/installed quickly. If that's not the case we can later revisit.

LaurentMazare:
Closing now that #55 has been released.

Issue #51, opened 2019-06-07
Shapes expect i64 instead of usize


grtlr:
Maybe this comes from generating the bindings automatically, but is there a use case for expecting `&[i64]` in `Tensor::reshape` and similar methods? Maybe it would make sense to change this to `&[usize]`? This change would also bring `Tensor` more in line with `ndarray`, which uses `usize` to represent shapes.

LaurentMazare:
Right, having `usize` at some places would be better. That said, a commonly used value is `-1` both for indexes and shapes.
For indexes it means start from the end, -1 would be the last value, -2 the value before and so on.
For shapes it's used for a single dimension and gets an appropriate value based on the tensor total number of elements (this number of element has to be divisible by the product of sizes for other dimensions of course). E.g. one could write `a.reshape([batch_size, -1])` to flatten `a`.
This explains why `i64` are used there, it's less verbose than using `Option<usize>` for shapes or some other alternative for indexes, it's also close to how the python api works but it certainly doesn't feel very rust idiomatic.

grtlr:
Thanks for the quick reply! I like the idea of `tch-rs` staying close to the Python API. For now I will just start to write some more idiomatic Rust wrappers around some of these methods for my own needs. Maybe in the future we can have a separate crate which handles these kind of things. Such a crate was briefly discussed in #12, do you have any ideas in mind here?

LaurentMazare:
I don't have much concrete ideas but maybe a couple of these would be good:
- The tensor type could be parameterized by the kind of element embedded would be nice, e.g. `Tensor<i64>`.
- The tensor type could embed the number of dimensions, e.g. `Tensor<d2, i64>` or `Tensor2d<i64>`.
- Maybe [tensors with named dimensions](http://nlp.seas.harvard.edu/NamedTensor) although that may be quite a stretch.

I would have hoped for the first two points to be doable when generating the binding code but I haven't found enough information in the `Declarations.yaml` file that is used for generation. Fwiw tensorflow provides some description of operations with all the details on the supported types which make it possible to generate better type annotations, e.g. for conv2d here is some [generated ocaml code](https://github.com/LaurentMazare/tensorflow-ocaml/blob/master/src/graph/ops_generated.mli#L2035-L2046): conv2d can only apply to tensors of float or double and return a tensor of the same type.

Also if you get some interesting wrappers that you think should be part of `tch-rs` let me know via a PR or an issue.

LaurentMazare:
[...]

PR #55, opened 2019-06-20
`TryFrom` and `TryInto` for `ndarray` types

grtlr:
This is a first implementation of #52. There still is an `unwrap()` in the `try_from` function which we should get rid of. Unfortunately, `?` does not seem to work with `std::option::NoneError` yet.

@LaurentMazare Do you have an idea how to fix this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ readme = "README.md"
 failure = "0.1"
 lazy_static = "1.3.0"
 libc = "0.2.0"
+ndarray = "0.12"
 rand = "0.6.5"
 torch-sys = { version = "0.1.0", path = "torch-sys" }
 zip = "0.5"
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -455,7 +455,7 @@ where
         // TODO: Replace this with `?` once it works with `std::option::ErrorNone`
         let slice = match value.as_slice() {
             None => failure::bail!("cannot convert to slice"),
-            Some(v) => v
+            Some(v) => v,
         };
         let tn = Self::of_slice(slice);
         let shape: Vec<i64> = value.shape().iter().map(|s| *s as i64).collect();
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -452,7 +452,12 @@ where
     type Error = Fallible<Tensor>;
 
     fn try_from(value: ndarray::Array<f64, D>) -> Result<Self, Self::Error> {
-        let tn = Self::of_slice(value.as_slice().unwrap());
+        // TODO: Replace this with `?` once it works with `std::option::ErrorNone`
+        let slice = match value.as_slice() {
+            None => failure::bail!("cannot convert to slice"),
+            Some(v) => v
+        };
+        let tn = Self::of_slice(slice);
         let shape: Vec<i64> = value.shape().iter().map(|s| *s as i64).collect();
         Ok(tn.reshape(&shape))
     }
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -449,7 +449,7 @@ impl<D> TryFrom<ndarray::Array<f64, D>> for Tensor
 where
     D: ndarray::Dimension,
 {
-    type Error = Fallible<Tensor>;
+    type Error = failure::Error;
 
     fn try_from(value: ndarray::Array<f64, D>) -> Result<Self, Self::Error> {
         // TODO: Replace this with `?` once it works with `std::option::ErrorNone`
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -1,6 +1,7 @@
 //! A Torch tensor.
 use crate::{Device, Kind};
 use failure::Fallible;
+use std::convert::{TryFrom, TryInto};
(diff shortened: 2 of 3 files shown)


LaurentMazare:
Indeed [NoneError](https://doc.rust-lang.org/std/option/struct.NoneError.html) will probably be the way to go when it gets to stable.
Meanwhile maybe you can do something very simple like the following?
```rust
    let slice = match value.as_slice() {
        None => bail!("cannot convert to slice"),
        Some(v) => v
    };
```
Alternatively you can also use `map_or` but that would probably be a bit less inefficient as I link it would create the error structure even on successes.

grtlr:
I am sorry, I don't really know the `failure` crate yet. The above code leads to the following error:

```
error[E0308]: mismatched types
   --> src/tensor/mod.rs:457:21
    |
457 |             None => failure::bail!("cannot convert to slice"),
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                     |
    |                     expected enum `std::result::Result`, found struct `failure::Error`
    |                     help: try using a variant of the expected type: `Err($crate::err_msg("cannot convert to slice"))`
    |
    = note: expected type `std::result::Result<wrappers::tensor::Tensor, failure::Error>`
               found type `failure::Error`
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

error: aborting due to previous error
```

LaurentMazare:
I think the culprit could be `type Error = Fallible<Tensor>;`. `Fallible<Tensor> is not really an error type but rather a [result or error type](https://docs.rs/failure/0.1.5/failure/type.Fallible.html).
Maybe something like `type Error = fallible::Error` would work ? I haven't tried though.

grtlr:
That was it, it should be working now.

LaurentMazare:
Looks good, thanks for your work on the PR.
It would be nice to support additional types so that we cover at least f64, f32, i64, i32. Maybe using a macro could be an easy way to do it, or maybe adding a type parameter to the added functions + a trait that is implemented by f64, f32, i64, i32.
Do you think you would want to try this in this PR ?
Alternatively I'm happy to merge the current version and either you and I can come back later on this.

grtlr:
I have been thinking a little bit on how to implement this. As far as I understand `Kind` is an enum, which specifies the different type of elements that a `Tensor` can hold, and comes directly from the underlying implementation.

Of course the straightforward way of dealing with different types would be just to pattern match on the different values of `Kind`. Arguably, the more idiomatic way to do this in Rust would be to have a `Kind` trait, on which `Tensor` is parameterized. Essentially that would probably mean that the pattern match would happen earlier in the wrapper. Have you thought about that case?

How should we proceed? If you decide to take the former route, I can implement that too.

LaurentMazare:
Indeed ideally we should have a type `Tensor<K>` for tensors where `K` implements some `Kind` trait.
The main reason why it's not the case is that most of the bindings code is generated automatically from a `Declarations.yaml` file that summarizes all the pytorch ops (this file itself is generated when compiling pytorch).
However I'm not sure that `Declarations.yaml` has enough type information for us to properly type all the operations. It contains some special tensor types, e.g. `IndexTensor` and `BoolTensor` which should help, but when using an ops like `equal` we probably want to check that the two arguments have the same kind and I don't really know how to infer that from the declarations file (the yaml section for this operator is at the bottom of this message). Maybe inferring that all the `Tensor` involved in the same op have the same kind and any kind is allowed would work but it's hard to be sure. I'll probably try to have a deeper look at this when I get a chance - or feel free to do so if you want, although the code generator is currently in ocaml.

Meanwhile I'm happy to go with the easy way for this PR. I would have thought that it's mostly a matter of using a macro so that we can create functions with the proper types and support `i32`, `i64`, `f32`, `f64`, but maybe matching on `Kind` would work too, happy to see a PR going either way.
(and again if you prefer getting this PR merged first then we work on the follow-up, happy to do so too)

```yaml
- name: equal
  method_prefix_derived: ''
  arguments:
  - dynamic_type: Tensor
    name: self
    type: const Tensor &
  - dynamic_type: Tensor
    name: other
    type: const Tensor &
  method_of:
  - Type
  - Tensor
  - namespace
  mode: TH
  buffers: []
  returns:
  - dynamic_type: bool
    name: result
    type: bool
  inplace: false
  abstract: true
  device_guard: true
  with_gil: false
  deprecated: false
```

grtlr:
> Meanwhile I'm happy to go with the easy way for this PR. I would have thought that it's mostly a matter of using a macro so that we can create functions with the proper types and support i32, i64, f32, f64, but maybe matching on Kind would work too, happy to see a PR going either way.
(and again if you prefer getting this PR merged first then we work on the follow-up, happy to do so too)

I'm happy with the macro route — I can add this to the current PR. Maybe I will find some time in the next days.

grtlr:
I have implemented a first version of the macros. Basically, `Tensor::of_slice`` does all the heavy lifting. I don't really know how to check whats going in Torch internally when we call this. Maybe you can review my implementation when you find time?

LaurentMazare:
Looks all good to me, thanks a lot for the PR!
The use of `usize` vs `i64` conversion is a bit sad as you mentioned in #51. It's probably a good time to revisit how `i64` are used in shape/index arguments.

PR #58, opened 2019-06-29
Derive Send and Sync

r-wheeler:
We are currently using this to load torch models in rust and are planning to wrap it in a multi threaded grpc server. Inspired from 

https://github.com/tensorflow/rust/blob/7a1700c50cb00c3cf4dcd0ffa438177a84e93a0d/src/session.rs#L251


LaurentMazare:
Thanks for the PR!

PR #60, opened 2019-07-07
Basic tensor indexing

jerry73204:
This PR aims to provide an `Tensor::i()` for convenient indexing. It covers integer (i64), Range*, slice of integers and new axis insertion. The following illustrates the usage.

```rust
use tch::{Tensor, IndexOp, NewAxis};

let tensor = Tensor::arange1(0, 2 * 3 * 5 * 7, opt).view(&[2, 3, 5, 7]);
let result = tensor.i((1, 1..2, vec![2, 3, 0].as_slice(), NewAxis, 3..)); 
```

So far, it allows basic indexing similar to that in [NumPy's indexing manual](https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.indexing.html). Advanced indexing schemes such as boolean array and multi-dimensional index array are not yet supported.

Note that in some cases the indexing result would slightly different NumPy's. For example:

```rust
let tensor = Tensor::zeros(&[3, 4, 5], opt);
let indexed = tensor.i(..1, vec![0, 3], vec![2, 1, 3]);
dbg!(indexed.size()); // result shape [1, 2, 3]
```

However, NumPy throws a shape mismatch error. According to NumPy's manual, multiple arrays as indexes will trigger advanced indexing. It expects `[0, 3]` and `[2, 1, 3]` have identical shape but actually not.
```python
# NumPy example
array[:1, [0, 3], [2, 1, 3]] // shape mismatch error
```

The NumPy promises that basic indexing always build the view of tensor, and I expect this PR respect this rule. In contrast, advanced indexing always makes a copy. That's one reason I decided to make it simple and put it apart.

### DIFF
--- a/src/tensor/index.rs
+++ b/src/tensor/index.rs
@@ -8,12 +8,12 @@ use std::ops::{
     RangeToInclusive,
     Bound,
 };
-use super::Tensor;
+use crate::Tensor;
 
-// https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html
-
+#[derive(Debug, PartialEq)]
 pub struct NewAxis;
 
+#[derive(Debug, PartialEq)]
 pub enum TensorIndexer {
     Select(i64),
     Narrow(Bound<i64>, Bound<i64>),
@@ -110,7 +110,7 @@ impl_from_range!(RangeInclusive<i64>);
 impl_from_range!(RangeTo<i64>);
 impl_from_range!(RangeToInclusive<i64>);
 
-trait IndexOp<T> {
+pub trait IndexOp<T> {
     fn i(&self, index: T) -> Tensor;
 }
 
--- a/src/tensor/index.rs
+++ b/src/tensor/index.rs
@@ -204,8 +204,21 @@ impl Tensor {
         use std::ops::Bound::*;
         use TensorIndexer::*;
 
+        // Make sure n. non-newaxis does not exceed n. of dimensions
+        let n_newaxis = index_spec
+            .iter()
+            .fold(
+                0,
+                |mut count, spec| {
+                    if spec == &InsertNewAxis {
+                        count += 1
+                    }
+                    count
+                }
+            );
+
         assert!(
-            index_spec.len() <= self.size().len(),
+            index_spec.len() <= self.size().len() + n_newaxis,
             format!("too many indices for tensor of dimension {}", self.size().len())
         );
 
@@ -213,8 +226,6 @@ impl Tensor {
         let mut curr_idx: i64 = 0;
 
         for (_spec_idx, spec) in index_spec.iter().enumerate() {
-            let dim_len = curr_tensor.size()[curr_idx as usize] as i64;
-
             let (next_tensor, next_idx) = match spec {
                 InsertNewAxis => (
                     curr_tensor.unsqueeze(curr_idx),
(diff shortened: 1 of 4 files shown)


jerry73204:
The lint rejects the coding style. Let me fix this.

src/tensor/index.rs
```diff
@@ -0,0 +1,304 @@
...
+        let d = index.3.into();
+        let e = index.4.into();
+        let f = index.5.into();
+        let g = index.6.into();
+        self.indexer(&[a, b, c, d, e, f, g])
+    }
+}
+
+impl Tensor {
+    fn indexer(&self, index_spec: &[TensorIndexer]) -> Tensor {
```
LaurentMazare: Similar to other tensor function, can you write a fallible version `f_indexer(...)` and have `indexer(...)` call it + `unwrap` ?

src/tensor/index.rs
```diff
@@ -0,0 +1,304 @@
...
+        use std::ops::Bound::*;
+        use TensorIndexer::*;
+
+        // Make sure n. non-newaxis does not exceed n. of dimensions
+        let n_newaxis = index_spec.iter().fold(0, |mut count, spec| {
```
LaurentMazare: What about something like `index_spec.iter().filter(|x| x == &InsertNewAxis).count()` ?

src/tensor/index.rs
```diff
@@ -0,0 +1,304 @@
...
+        // Apply indexing from left to right
+        let mut curr_tensor = self.shallow_clone();
+        let mut curr_idx: i64 = 0;
+
+        for (_spec_idx, spec) in index_spec.iter().enumerate() {
+            let (next_tensor, next_idx) = match spec {
+                InsertNewAxis => (curr_tensor.unsqueeze(curr_idx), curr_idx + 1),
+                Select(index) => (
+                    curr_tensor.select(curr_idx, *index),
+                    curr_idx, // not advanced because select() sequeezes dimension
```
LaurentMazare: sequeezes -> squeezes ?

src/tensor/index.rs
```diff
@@ -0,0 +1,304 @@
...
+                "too many indices for tensor of dimension {}",
+                self.size().len()
+            )
+        );
+
+        // Apply indexing from left to right
+        let mut curr_tensor = self.shallow_clone();
+        let mut curr_idx: i64 = 0;
+
+        for (_spec_idx, spec) in index_spec.iter().enumerate() {
```
LaurentMazare: If `_spec_idx` is not used, maybe don't use `enumerate()` ?

tests/tensor_indexing.rs
```diff
@@ -0,0 +1,114 @@
...
+use tch::{IndexOp, NewAxis};
+
+#[test]
+fn integer_index() {
+    let opt = (Kind::Float, Device::Cpu);
+
+    let tensor = Tensor::arange1(0, 2 * 3, opt).view(&[2, 3]);
+    let result = tensor.i(1);
+    let expect = Tensor::arange1(3, 6, opt);
+    assert!(result.eq1(&expect).all().int64_value(&[]) == 1);
```
LaurentMazare: When comparing the result tensor to a 1-d tensor, I find the following to be slightly more readable:
`assert_eq!(Vec::<i64>::from(&result), [3, 4, 5]);`
(the same applies to further tests below)

LaurentMazare: I would also add an `assert_eq!(result.shape, [3])` just in case.
(the same applies to further tests below, maybe when the result is not a 1d tensor you can assert both on shape + on the flattened version?)

src/tensor/index.rs
```diff
@@ -0,0 +1,304 @@
+use crate::Tensor;
```
LaurentMazare: Consider including some module level documentation including a couple examples ?

src/tensor/index.rs
jerry73204: It was left for (maybe) advanced indexing in the future. I remove it for now.

src/tensor/index.rs
jerry73204: Good idea. I don't have spare time until this weekend. You may merge the PR and I'll write the doc later, or wait for it to happen.

LaurentMazare:
Merged, thanks again for the PR. I'll certainly use this in new models and hopefully adapt some of the existing examples.

PR #68, opened 2019-07-14
Implement MobileNetV3 and examples

jerry73204:
This is my attempt to merge my work to tch. I introduces large and small versions of model along with an complete example.


jerry73204:
Update some progress here. I saw a relevant [thread](https://stats.stackexchange.com/questions/272607/cifar-10-cant-get-above-60-accuracy-keras-with-tensorflow-backend) on SO. In brief, some fine-tuning and augmentation would help.

The vanilla settings (not shuffle, no augmentation) will stuck in low accuracy, even you lengthen the # of epochs. I've started running experiments with augmentation, and I observed obvious improvement. (So far 87% on validation set.) Let's see if it finally reaches paper's claim.

I changed the example a bit to adopt existing shuffling and augmentation in tch. By the way, we could leave explicit comments on fine-tuning at the end to help new-comers.

LaurentMazare:
Nice to hear that, hopefully it will improve a bit more on top of this (my experience with densenet was that it reaches 93% test accuracy without augmentation and almost 95% test accuracy with).

jerry73204:
After 1000 epochs of training,v2 converges at 85% and v3 reaches 90% (and of course, 99% training accuracy). I think this PR would hold for a while before we find the _sauce_. Now I'm running other's Keras/TF impl to see if there's any differences.

LaurentMazare:
1000 epochs is a lot, my experience would be that it should converge in quite less than this. Looking quickly at your code, it seems that you use a constant learning rate + adam ? I think dividing the learning rate by 10 after say 50 epochs and 100 epochs would improve things significantly, also a properly tuned sgd may slightly outperform adam.
Maybe you could have a look at `examples/cifar/main.rs`, this reaches test
`~95%` accuracy in 150 epochs. You can clearly see the effect of the learning rate drop at epochs 50 and 100 on the plot below (this runs in a bit less than a hour on a P100).
image

jerry73204:
I knew lr schedule is a thing, but does not know the significance. I'll give it a try.

jerry73204:
It's long time and I'm back. I took a glance at [pytorch-cifar](https://github.com/kuangliu/pytorch-cifar) and [mobileNet-v2_cifar10](https://github.com/tinyalpha/mobileNet-v2_cifar10) that claims 94% accuracy, and went for the detour to fix memory leaking issue. Let me made some conclusions here.

### MobileNetV2
The reference impls [pytorch-cifar](https://github.com/kuangliu/pytorch-cifar) and [mobileNet-v2_cifar10](https://github.com/tinyalpha/mobileNet-v2_cifar10) don't follow the exact model on paper. They reduces strides on some layers that increases the whole capacity. I did the same on ours and yes, it finally reaches test accuracy of 93% (I think 94% is the peak value).

In summary, the working settings are expanded capacity, and SGD with lr {0.1, 0.01, 0.001} on epochs {50, 100, 150}. So current impl would be fine.

### MobileNetV3
There are two known impls [pytorch-mobilenet-v3
](https://github.com/kuan-wang/pytorch-mobilenet-v3) and [mobilenet_v3](https://github.com/SpikeKing/mobilenet_v3). I double checked the paper and conclude they are correct impls. I ran the both on cifar-10 and stuck at training acc 85% and testing acc 80%., and so do ours. It does not improve which ever dropout and channel multiplier you choose (paper suggests 0.8 dropout, and I tested 0.5 and 0.0, and up to 3 mutiplier).

Interestingly, if you copy the param settings from paper (RmsProp + EMA + lr schedule with some adjustment wrt to batch size). Things just get worse. In fact, in the paper, the RmsProp has unreasonably large momentum 0.9, while TF and Torch defaults to 0.0. Ours falls into NaNs with large momentum. Even you reduce it, the acc never crosses 50% if EMA is present. Therefore, I decided to ignore it anyway.

@LaurentMazare how do you think of this?

jerry73204:
Plus, both the master branch on CI and mine cannot build.

LaurentMazare:
> Plus, both the master branch on CI and mine cannot build.

Yes that's probably because we switched to PyTorch 1.2. Updating should get rid of these errors. As for CI this will be broken until the pre-compiled packages for Linux have been released (which somehow is not the case yet).
This issue is tracked here: https://github.com/LaurentMazare/tch-rs/issues/91
Sorry about this, hopefully everything should be fixed soon.

LaurentMazare:
Closing as this is probably a bit out of date now, feel free to re-open if you want this model to be added.

PR #77, opened 2019-07-21
Bug fix: undefined reference to ato_rms_prop

jerry73204:
This is due to inconsistent names in `torch_api.h` (`ato_rmsprop`) and in `wrappers/optimizer.rs` (`ato_rms_prop`).


LaurentMazare:
Thanks for catching this and fixing it!

PR #78, opened 2019-07-22
Implement exponential moving average

jerry73204:
It is required feature when reproducing MobileNetV3. The TensorFlow equivalence is [tf.train.ExponentialMovingAverage](https://www.tensorflow.org/api_docs/python/tf/train/ExponentialMovingAverage).

It keeps shadow copies of trainable variables from a VarStore. When `step()` is called, it applies weighted average by `shadow_variable -= (1 - decay) * (shadow_variable - variable)` and copies the shadow copy back to trained variable.

### DIFF
--- a/src/nn/ema.rs
+++ b/src/nn/ema.rs
@@ -1,0 +1,71 @@
+//! Exponential moving average.
+use super::var_store::VarStore;
+use crate::Tensor;
+
+/// A ExponentialMovingAverage maintains moving averages of the trainable variables.
+///
+/// It keeps shadow copies of trainable variables in VarStore. In each step,
+/// it applies average by `shadow_variable -= (1 - decay) * (shadow_variable - variable)`,
+/// then copies updated shadow variables back to VarStore.
+#[derive(Debug)]
+pub struct ExponentialMovingAverage {
+    decay: f64,
+    shadow_target_vars: Vec<(Tensor, Tensor)>,
+}
+
+impl ExponentialMovingAverage {
+    /// Constructs a new ExponentialMovingAverage instance.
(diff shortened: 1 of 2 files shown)


jerry73204:
In addition, I noticed `+-*/` operators does not implement on `&mut Tensor`. It leads the inconvenience of `&*tensor` conversion. Consider open another issue for this?

LaurentMazare:
Do you know if there is something similar in the PyTorch python api ? I haven't found anything after some quick googling so I'm wondering why/if there is no support for this. If there is a PyTorch version, we should probably follow it closely so that it's easier for python users to start using rust.
Re the `&*tensor` that's indeed a bit sad (and afaik is more of a rust culprit). Having the binary ops `+`, `-`, `*`, `/`, work on `&mut Tensor` would sound reasonable to me.

jerry73204:
I grep whole my local pytorch doc and see no clue about the support. I agree that we should keep close to pytorch, but it's officially supported. The work was part of my effort to reproduce MobileNetV2 experiment since it was mentioned in MNetV{2,3} and inception paper. I think it could be shared to public.

I'm not sure if it's suitable to put it into tch, or something like `tch-contrib`, or I maintain my own. They are acceptable if we make the rule explicit.

LaurentMazare:
If your plan is to have mobilenet-v3 in examples/ maybe we can start by having the ema module there too. This way it would be easy for users to find it and if it starts being useful to other models we can move it to `src/nn` or some other common place.
Would that seem reasonable to you?

jerry73204:
Sounds good. Let me close this PR.

PR #80, opened 2019-07-22
Revise tensor indexing doc

jerry73204:
Fix a typo and put some notice that the indexing result maybe different from NumPy. Besides, I found the whole module doc is not compiled into HTM since the `tch::tensor::index` module is not public. Consider to make a fix?

### DIFF
--- a/src/tensor/index.rs
+++ b/src/tensor/index.rs
@@ -33,7 +33,7 @@
 //! assert_eq!(Vec::<i64>::from(t.contiguous().view(-1)), [2, 3, 5, 6]);
 //! ```
 //!
-//! The `NewAxis` index can be used to insert a dimensio.
+//! The `NewAxis` index can be used to insert a dimension.
 //!
 //! ```
 //! use crate::tch::{Tensor, IndexOp, NewAxis};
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Thanks for the PR.
How did you generate the html help ? (if you looked at docs.rs and found it not to include index, that would be expected as I haven't made a release since then)

jerry73204:
Simply `cargo doc --release`. I checked `grep -ri numpy target/doc/tch/` and found nothing.

LaurentMazare:
Thanks, I just ran this command, and in `target/doc/src/tch/tensor/index.rs.html` I get to see your newly added numpy comments. Maybe it didn't refresh for you for some reason?

jerry73204:
I checked `target/doc/src` and they are docs of _source code_. I expect a single module page in `target/doc/tch`. Is it fine for you?

LaurentMazare:
You're right, it's indeed the formatted source code, I should have looked at the document rather than just grepping.
Anyway I just exposed the `index` module at top level and the command now produces `target/doc/tch/index/index.html`.

LaurentMazare:
I just updated the crate package and the doc can now be seen on docs.rs, thanks for all your work on this!
https://docs.rs/tch/0.1.1/tch/index/index.html

PR #96, opened 2019-08-19
Build and link libtorch with MSVC

jerry73204:
It relates to issue #9 that the PR builds against MSVC and links to libtorch. Windows specific instructions are made into README.


jerry73204:
I don't expect CI will cover this. So here shows my successful build in screencast.
https://drive.google.com/file/d/1V7a5obWQ1cfPlVkrSnHUTx2GdFktrxvH/view?usp=sharing

LaurentMazare:
Thanks for the PR! (I wouldn't be able to test the windows build but if it works for you this is already better than the current state).

PR #97, opened 2019-08-19
torch-sys refactoring

### LINKED ISSUE #95: Linking error tch 0.1.1 & Centos 7
When I try to run a binary using `tch = 0.1.1` on `centos 7`, I get a linking error:
`/usr/bin/ld: warning: libgomp-8bba0e50.so.1, needed by /home/maxence/zezima/market-analytics/target/release/build/torch-sys-c846cc9c89b11e6f/out/libtorch/libtorch/lib/libc10.so, not found (try using -rpath or -rpath-link)`. Note that I'm not installing libtorch manually.

When I look at the content of the `OUT_DIR` of the build script, I can see `libgomp-8bba0e50.so.1`. So the lib is not missing, just not linked.

See the full error output https://gist.github.com/jean-airoldie/c908704722181f2d0dfc27bf64bcf668.

jean-airoldie:
~~Closes #95.~~

~~If windows is officially supported, we might want to add a `appveyor.yml` for CI and test against that before merging (#98).~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,11 @@ zip = "0.5"
 
 cpython = { version = "0.2.0", optional = true }
 
+[workspace]
+members = [
+    "torch-sys",
+]
+
 [features]
 python = ["cpython"]
 
--- a/torch-sys/Cargo.toml
+++ b/torch-sys/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.2"
 authors = ["Laurent Mazare <lmazare@gmail.com>"]
 edition = "2018"
 build = "build.rs"
-links = "torch"
+links = "tch"
 
 description = "Low-level FFI bindings for the PyTorch C++ api (libtorch)."
 repository = "https://github.com/LaurentMazare/tch-rs"
@@ -17,7 +17,6 @@ libc = "0.2.0"
 
 [build-dependencies]
 cc = "1.0"
-cmake = "0.1"
 curl = "0.4.9"
 failure = "0.1"
 zip = "0.5"
--- a/torch-sys/build.rs
+++ b/torch-sys/build.rs
@@ -5,19 +5,13 @@
 //
 // On Linux, the TORCH_CUDA_VERSION environment variable can be used,
 // like 9.0, 90, or cu90 to specify the version of CUDA to use for libtorch.
-#[macro_use]
-extern crate failure;
-
-use std::env;
-use std::fs;
-use std::io;
-use std::io::Write;
-use std::path::{Path, PathBuf};
-
-use cmake::Config;
 use curl::easy::Easy;
-use failure::Fallible;
-use zip;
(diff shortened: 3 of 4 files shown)


jean-airoldie:
Also you might want to rename `torch-sys` to `tch-sys` since the compiled static lib we link against is called `libtch`.

jean-airoldie:
Should the `torch-sys/src/c_generated.rs` be added to the gitignore?

LaurentMazare:
> Should the `torch-sys/src/c_generated.rs` be added to the gitignore?

I don't think so, these generated files are in the repo so that users of the crate don't need to run the code generation when compiling the library (as this bit is written in ocaml). Also these files should only change when supporting a new version of PyTorch, e.g. when we moved from PyTorch 1.1.0 to PyTorch 1.2.0

jean-airoldie:
Is the code generated externally from this repo? Because if they are generated in a `build.rs` script, we should add a feature to regenerate bindings.

LaurentMazare:
It is not generated from `build.rs`.

jean-airoldie:
Alright. I'm not sure why I am credited with change the `c_generated.rs`. I might have rebased on something I should not have.

jean-airoldie:
I removed the cmake dependency, because I didn't really see the point.

LaurentMazare:
> I removed the cmake dependency, because I didn't really see the point.
I use it in some setting as it's closer to what is used to compile libtorch C++ programs.

Overall this PR seems to make lots of orthogonal changes which makes it difficult to review.
Does the fix for #95 boil down to adding `-lgomp` and renaming the generated library from `libtorch` to `tch` ? If so maybe we can have a PR that does just this and discuss other changes in another PR. Also are both bits necessary or is one of them enough?

jean-airoldie:
> I use it in some setting as it's closer to what is used to compile libtorch C++ programs.

Not sure I understand what that means.

> Does the fix for #95 boil down to adding -lgomp and renaming the generated library from libtorch to tch ?

It might be possible to fix #95 without having to rename. I can make a PR exclusively for that and relabel the present PR as a build script refactoring.

jean-airoldie:
Alright so here is a better description of this PR.

Currently both the `cmake` and `cc` build a static library called `liblibtorch` which is linked against `libtorch`. However this `liblibtorch` isn't actually used Instead we link against the precompiled `libtorch` shared lib. This is why `-Wl,-rpath` is needed to properly link the lib at runtime.

This PR instead compile a static library called `libtch` which is link against `libtorch` and all its dependencies. It is this `libtch` that is exposed by `torch-sys`, so no relative path shenanigans are needed since `libtch` knows the location of the shared libs at runtime.

LaurentMazare:
Sorry for letting this hang forever, I think the important bit of this PR (renaming liblibtorch to libtch) is now part of the build setup, so closing this for now.

PR #98, opened 2019-08-19
Added appveyor.yml

jean-airoldie:
Assuming that windows is officially supported, here is an appveyor.yml for rust.


LaurentMazare:
Thanks. I'm not sure about 'official' windows support but we can enable this, see how it goes, and revisit if there are some issues.

Issue #123, opened 2019-11-23
Heap gets corrupted with LSTM models in latest versions.


svenslaggare:
The heap gets corrupted when using an exported LSTM model. It started to happening on the most recent commits. Using commit '81f6215c8f7a374b89591abbbd61ac3bfe24dc14' does _not_ lead to a corrupted heap.

When running the example below, the program crashes with "double free or corruption".

I'm using:
- Ubuntu 18.04
- Rust 1.39

Model definition (python):
```
import os
import numpy as np

import torch
from torch import nn

class PredictionModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.lstm = nn.LSTM(3, 200, num_layers=1)
        self.output = nn.Linear(200, 3)

    def forward(self, x, hidden=None, lengths=None):
        y, hidden = self.lstm(x, hidden)
        y = self.output(y)
        return y, hidden

    def hidden_zero(self, batch_size):
        return \
            torch.zeros(1, batch_size, self.lstm.hidden_size), \
            torch.zeros(1, batch_size, self.lstm.hidden_size)

if __name__ == "__main__":
    model = PredictionModel()
    model.eval()

    with torch.no_grad():
        inputs = torch.FloatTensor(np.zeros((1, 1, 3)))
        hidden = model.hidden_zero(1)

        traced_model = torch.jit.trace(model, (inputs, hidden))
        traced_model.save(os.path.join("compiled_model.pt"))
```

Model usage (rust):
```
use tch::{IValue, Tensor, CModule, Kind};

fn main() {
    let model = CModule::load("compiled_model.pt").unwrap();

    let options = (Kind::Float, tch::Device::Cpu);
    let mut state = IValue::Tuple(vec![
        IValue::Tensor(Tensor::zeros(&[1, 1, 200], options)),
        IValue::Tensor(Tensor::zeros(&[1, 1, 200], options)),
    ]);

    let mut input_tensor = IValue::Tensor(Tensor::zeros(&[1, 1, 3], options));
    let mut prediction = model.forward_is(&[
        &input_tensor,
        &state
    ]).unwrap();
}
```

svenslaggare:
I think I have found the problem and the solution.

[...]

PR #124, opened 2019-11-23
Properly free IValue::tuple elements (issue #123).

svenslaggare:
The 'ati_deep_free' function is removed as it causes heap corruption. The IValue::to_c function has been modified to avoid the memory leaks that the 'ati_deep_free' function tried to solve.

See more details in the issue https://github.com/LaurentMazare/tch-rs/issues/123.


LaurentMazare:
Cool, thanks for investigating and fixing this!

PR #134, opened 2019-12-29
Fix the bug in constant weight initializer

pkubik:
In `init` function the branch for `Init::Const` contains an optimization for zeroes and ones initialization. "Ones "optimization is supposed to happen when `cst` is close to 1 but it was triggered when `cst` was just lower than 1. Fixed by adding `abs()` to the expression.


LaurentMazare:
Thanks for catching this, I'll add a test for this case so that this won't happen again.

PR #135, opened 2020-01-03
Add example `custom-optimizer` with Adam optimizer

bytesnake:
This example demonstrates how a custom optimizer can be implemented in Rust. This Adam optimizer supports sparse, as well as, dense gradient updates and can be used in problems, where only a small portion of the training parameters are updated. (e.g. embedding matrices)

The toy example is a modified version of the MNIST-NN example. It reaches 97% in about 170 epochs for both sparse and dense mode.


LaurentMazare:
Looks very nice, thanks for the PR!

PR #137, opened 2020-01-07
More tensor conversions from Vec and primitives

jerry73204:
The PR made following changes.

- Implement `From<Vec<_>> for Tensor`
- Support array or vec of bool type conversion
- More conversion tests

### DIFF
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -560,6 +560,7 @@ try_into_impl!(f32);
 try_into_impl!(i32);
 try_into_impl!(f64);
 try_into_impl!(i64);
+try_into_impl!(bool);
 
 macro_rules! try_from_impl {
     ($type:ident) => {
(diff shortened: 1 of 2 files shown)


src/tensor/mod.rs
```diff
@@ -575,15 +576,25 @@ macro_rules! try_from_impl {
                     None => failure::bail!("cannot convert to slice"),
                     Some(v) => v,
                 };
-                let tn = Self::of_slice(slice);
+                let tn = Self::f_of_slice(slice)?;
                 let shape: Vec<i64> = value.shape().iter().map(|s| *s as i64).collect();
                 Ok(tn.reshape(&shape))
```
LaurentMazare: I cannot give it a try right now but maybe it would make sense to even use `f_reshape` here?

jerry73204: Sounds good. Change it accordingly.

LaurentMazare:
Merged, thanks for the PR!

jerry73204:
I noticed `From<Vec<Vec<_>>>` is not implemented for `Tensor`. If so, the vector length sanity check will be done in `try_from()`. How do you think about this?

LaurentMazare:
Sorry I just came across your last message. Would that be an issue? Failing in the try_from should return an appropriate error hopefully (but it may be good to add a test case for this).

jerry73204:
The `Vec<Vec<_>>` type has no content of dimension size. I was concerning the case that `Tensor` size is erroneously calculated in some cases and the conversion to `Vec<Vec<_>>` will silently get the wrong size.

Maybe I worry too much. I can send a PR and implement as it is and place the dimension verification issue on other thread.

PR #138, opened 2020-01-10
Make normalize and unnormalize functions public.

de-vri-es:
This PR makes `imagenet::normalize` and `imagenet::unnormalize` public, so they can be used on data acquired from something other than `imagenet::load_image`.

The use case we're thinking about specifically is running a network on data retrieved directly from a camera.


LaurentMazare:
Merged! Running from a camera directly would be a pretty cool application!

de-vri-es:
Awesome, thanks!

PR #140, opened 2020-01-14
Fix Init::Randn with a std dev (0, 1) and zero mean

danieldk:


LaurentMazare:
Thanks for catching this issue and fixing it (and adding a test).

PR #148, opened 2020-01-26
Extended Embedding configuration and added unit tests

guillaume-be:
The padding index was set to a value of -1 for all embedding layers created. Being able to configure this value is useful when working with pretrained models. Added this field to the EmbeddingConfig (keeping the default to -1). Added a few unit tests for the embedding layer that were missing.


LaurentMazare:
Thanks for the PR!

PR #155, opened 2020-02-08
Renamed embeddings weight variable to be aligned with Torch

guillaume-be:
The default Embedding implementation in Pytorch uses a parameter variable named "weight" to store the embedding matrix (aligned with the rest of the base modules). https://pytorch.org/docs/stable/_modules/torch/nn/modules/sparse.html#Embedding

Proposing to change the current "embedding" variable name to facilitate loading of pre-trained models from Pytorch.

I looked for use in the current code base, but as far as I could see, no pre-trained embedding layers were loaded so far.

### DIFF
--- a/src/nn/sparse.rs
+++ b/src/nn/sparse.rs
@@ -44,7 +44,7 @@ pub fn embedding<'a, T: Borrow<super::Path<'a>>>(
     let vs = vs.borrow();
     Embedding {
         ws: vs.var(
-            "embedding",
+            "weight",
             &[num_embeddings, embedding_dim],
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Seems fine, thanks for the PR.
Could you try running rustfmt, travis CI seems to be broken because of this.

guillaume-be:
Yes - sorry about that I thought I could outsmart rustfmt by just changing the value of this string. Lesson learned :)

LaurentMazare:
Merged, thanks again.

PR #156, opened 2020-02-08
Allow loading partial sets of pretrained weights

guillaume-be:
The current load method for loading weights in the var store performs a robust check that all of the variables of the var store can be found in the weights file. This is adequate for running inference or fine-tuning without architecture change.

It is common to load a set of pre-trained weights for the base of a model (e.g. conv blocks, embeddings, language models) and stack a few untrained layers on top of this pre-trained base. These layers are commonly trained from scratch.

Proposing adding a load_partial method that would allow loading a subset of the weights, returning a message with the weights that were not updated instead of panicking.

### DIFF
--- a/src/nn/var_store.rs
+++ b/src/nn/var_store.rs
@@ -143,7 +143,7 @@ impl VarStore {
     /// weight for only parts of the model are available.
     /// Note that the set of variables stored in the var-store is not changed, only the values
     /// for these tensors are modified.
-    pub fn load_partial<T: AsRef<std::path::Path>>(&mut self, path: T) -> Fallible<()>  {
+    pub fn load_partial<T: AsRef<std::path::Path>>(&mut self, path: T) -> Fallible<()> {
         let named_tensors = Tensor::load_multi_with_device(&path, self.device)?;
         let named_tensors: HashMap<_, _> = named_tensors.into_iter().collect();
         let mut variables = self.variables_.lock().unwrap();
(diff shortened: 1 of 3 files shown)


src/nn/var_store.rs
```diff
@@ -134,6 +135,33 @@ impl VarStore {
...
+    pub fn load_partial<T: AsRef<std::path::Path>>(&mut self, path: T) -> Fallible<()> {
+        let named_tensors = Tensor::load_multi_with_device(&path, self.device)?;
+        let named_tensors: HashMap<_, _> = named_tensors.into_iter().collect();
+        let mut variables = self.variables_.lock().unwrap();
+        for (name, var) in variables.named_variables.iter_mut() {
+            match named_tensors.get(name) {
+                Some(src) => {
+                    crate::no_grad(|| var.f_copy_(src).map_err(|e| format_err!("{}: {}", name, e)))?
+                }
+                None => println!(
```
LaurentMazare: Printing to stdout seems a bit fragile, maybe we could instead return the variable names that were not updated and it would be up to the caller to decide what do it with it?

guillaume-be: Thank you - this is a good idea. Now returning an optional vector of Strings with the missing variable names. Not sure if a consistent (and potentially empty vector) or an Optional vector would be more fitting the style of the crate.

LaurentMazare: Nice, thanks actually I think I would slightly lean towards a potentially empty vector rather than an optional vector as it seems a bit simpler (and this way no need for the caller code to assume that `Some(v)` always means that `v` is non empty).
Also could you update the doc comment of the function to explained the returned value.

guillaume-be: Ok - I will update to always return a vector, and will add the return type to the documentation. I also noticed that while unit tests run fine on my local set-up, they do fail on the Travis CI. I believe the reason is a race condition on the files written by the 4 unit tests.

I could see that each file name should have the ProcessId in it - not sure how Rust manages the multithreading for running tests. I see two potential solutions:
a) run the unit tests on a single thread (--test-threads=1)
b) hardcode a different file name for each of these var stores
c) Try replacing ProcessId by std::thread::ThreadId in the filename and see if it solves the issue

LaurentMazare: I think (b) is probably the easiest for now (and probably what we already do for some other tests).

LaurentMazare:
Thanks for the PR!

Issue #157, opened 2020-02-09
Consider tracking Declarations.yaml in git


de-vri-es:
I was planning to make some changes to code generation to work around #146. However, it appears I would need to compile pytorch manually to obtain a usable `Declarations.yaml` file.

I think it could make sense to track the `Declarations.yaml` file that was used for code generation. Then, anyone can make changes to the code generation and re-run it without compiling the right version of pytorch manually. That would save a lot of time and effort.

Do you think this would make sense?

LaurentMazare:
It would certainly be helpful to have this file tracked in the git repo. The main reason why I haven't added it is that I don't really know what license would apply to it and how ok that would be with the current tch-sys double licensing (apache/mit).
If you have any pointers that makes you think this would be ok, I would be happy to add the file!

de-vri-es:
I think it would be fine to just add the file and clarify somehow that a difference license applies to that file. It could be done by embedding the license in the file, or by adding the license right next to the file under a descriptive name (maybe `Declarations.yaml.LICENSE`). The second option has the advantage that the file can be copied verbatim every time you want to update it.

The yaml file doesn't end up in binary releases, so it only needs to be transferred with source copies. That's why I think it would be fine if the license file isn't at the root of the repository. If the license did have to be copied with binary releases, it would be too easy for packagers to miss the license in a subdirectory.

Another option could be to add a file called torch.LICENSE with a note at the top that explains for which files that license applies.

LaurentMazare:
Closed via some variant of #160

PR #158, opened 2020-02-09
Allow building with FFI stubs without linking to torch.

### LINKED ISSUE #146: Docs not built for version 0.1.3 (latest version as of 23-01-2020)
The Doc build on Doc.rs has failed for this release. 
The error is : 
```
[INFO] [stderr]    Compiling torch-sys v0.1.3
[INFO] [stderr] error: failed to run custom build command for `torch-sys v0.1.3`
[INFO] [stderr] 
[INFO] [stderr] Caused by:
[INFO] [stderr]   process didn't exit successfully: `/opt/rustwide/target/debug/build/torch-sys-1d88143c231e6ea2/build-script-build` (exit code: 101)
[INFO] [stderr] --- stderr
[INFO] [stderr] thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { description: "Couldn\'t resolve host name", code: 6, extra: None }', src/libcore/result.rs:1189:5
[INFO] [stderr] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
``` 

Looks like a dependency (`torch-sys`) has failed to build. 
Could we retry the build or something of that sort?

de-vri-es:
This PR makes it possible to build the `tch-sys` crate with stubs that don't link against `libtorch`. It is mainly intended for building the documentation without `libtorch`. This PR should fix #146.

The stubs are used when the `real-ffi` feature is disabled, which is enabled by default. The stubs simply panic with a descriptive message when called.

I think enabling the real FFI functions with a default feature is better than disabling them with a non-default feature. This allows us to avoid downloading dev dependencies when `real-ffi` is disabled, and it means that `--all-features` doesn't produce an unusable library. Instead, `--no-default-features` produces an unusable library, but I think that's better than vice-versa.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,8 +26,13 @@ cpython = { version = "0.2.0", optional = true }
 members = ["torch-sys"]
 
 [features]
+default = ["real-ffi"]
 python = ["cpython"]
+real-ffi = ["torch-sys/real-ffi"]
 
 [[example]]
 name = "reinforcement-learning"
 required-features = ["python"]
+
+[package.metadata.docs.rs]
+no-default-feautres = true
--- a/gen/gen.ml
+++ b/gen/gen.ml
@@ -567,17 +567,16 @@ let write_ffi funcs filename =
       pm "use crate::{C_scalar, C_tensor};";
       pm "use libc::c_int;";
       pm "";
-      pm "extern \"C\" {";
       Map.iteri funcs ~f:(fun ~key:exported_name ~data:func ->
           match func.Func.returns with
           | `fixed _ ->
-              pm "    pub fn atg_%s(out__: *mut *mut C_tensor, %s);"
+              pm "    pub_ffi!(fn atg_%s(out__: *mut *mut C_tensor, %s));"
                 exported_name
                 (Func.c_rust_args_list func)
           | `dynamic ->
-              pm "    pub fn atg_%s(%s) -> *mut *mut C_tensor;" exported_name
+              pm "    pub_ffi!(fn atg_%s(%s) -> *mut *mut C_tensor);" exported_name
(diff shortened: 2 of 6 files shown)


de-vri-es:
Judging from the amount of deleted lines from `c_generated.rs`, I did not have the right `Declarations.yaml` file when running the the code generation.

Weirdly enough, I didn't get any changes with respect to master when I ran the code generation without modifications.

@LaurentMazare: Could you perhaps share the `Declarations.yaml` file that should be used, or even track it in git (#157)?

de-vri-es:
It's not a wrong `Declerations.yaml`. Since all the declarations are now inside a macro invocation, `rustfmt` isn't formatting them anymore, so every function is now on a single line.

We could make the code generation insert line breaks for readability, if you like.

LaurentMazare:
Thanks for giving a try at fixing this. It turns out that I pushed a fix which needs a smaller change (see this [commit](https://github.com/LaurentMazare/tch-rs/commit/e91168570a05aade3db37116647fa26ba7da00b6) for details). The documentation on docs.rs is now up to date!

de-vri-es:
Ah nice. I suppose you would get linking errors with that, but if `cargo doc` doesn't do linking then that doesn't matter :)

Using a feature that disables linking does mean that builds with `--all-features` break though. I don't think that's the end of the world, but it might be a reason to consider inverting the meaning of feature and adding it to the default features.

Either way, glad there is a fix. I'll close this PR :)

de-vri-es:
After some more thinking: would it make sense to use `#[cfg(doc)]` instead of a feature? I suspect that with `#[cfg(doc)]`, dependent crates don't need to repeat the same trick to get their docs to build.

PR #160, opened 2020-02-15
Track Declarations.yaml including license.

### LINKED ISSUE #157: Consider tracking Declarations.yaml in git

de-vri-es:
This PR tracks `Declarations.yaml` and the covering license. This should make it easier for people to re-run code generation without compiling `libtorch` from source.

Let me know if you prefer a different scheme for including the license.

Would close #157


LaurentMazare:
Thanks for the PR, I just pushed something very equivalent to it (but didn't want to make my changes on top of yours as the yaml files are pretty big so I would prefer reducing the diff on them to as little as possible).
Difference with yours:
- The `Declarations.yaml` file is slightly different in the sense that each function definition starts with the `name` field, this ended up being useful while debugging the code generation as it's easier to grep/extract a single function definition.
- I've put the files in `third_party/pytorch` as there was no good reason for it to be in `data` in the first place.
- A small README has been added.

de-vri-es:
Cool, thanks for the update :)

PR #161, opened 2020-02-15
Expose (get|set)_num(_interop)?_threads

danieldk:
I need to set the number of threads in (as a command-line option) of an application, rather than through the `OMP_NUM_THREADS` environment variable. So, I thought it might be useful to expose ATen's functions for getting/setting the number of threads.


LaurentMazare:
Looks all good, thanks for the PR!

PR #164, opened 2020-03-02
Expose Tensor::data_ptr.

pbridger:


LaurentMazare:
Thanks for the PR, the Travis CI currently fails because of formatting. Could you please run rustfmt on the modified files ? Thanks!

pbridger:
Sure will. I am still a rust noob, so thanks for the guidance!

torch-sys/libtch/torch_api.h
```diff
@@ -130,6 +131,7 @@ ivalue atm_forward_(module,
                     ivalue *ivalues,
                     int nivalues);
 void atm_free(module);
+void atm_to(module m, at::Device device, at::ScalarType dtype, bool non_blocking);
```
LaurentMazare: You should not use C++ syntax in this file but only pure C (as it could be compiled by a C compiler).
In order to pass a scalar type, see how this is done in `at_tensor_of_data.`.

pbridger: Done, good point.

LaurentMazare:
Merged, thanks for the PR!

PR #170, opened 2020-03-05
added constructor for linear layer without bias

guillaume-be:
The bias in the original Linear layer from Pytorch is optional and may be skipped (https://pytorch.org/docs/stable/_modules/torch/nn/modules/linear.html#Linear)

A few architectures do not use a bias for specific linear layer. For example are the linear layers with weight tied to the input embeddings for encoder/decoder architectures or masked language models need to be equivalent to a reverse embedding layer and therefore do not use a bias (e.g. GPT2).

The current API forces the bias to be register to the varstore. This PR proposes a constructor for the linear layer that would allow creating linear layers without bias.
- adds a `bias` boolean parameter
- when set to false, the Linear layer bias is a tensor of zeros not registered to the var store

This is meant to maximize compatibility with current code and use a single Linear layer. Alternatives would be :
- set the bias in the linear layer to be of type Option<Tensor>, potentially breaking existing code accessing the bias Tensor directly
- create a new LinearWithoutBias module that would duplicate a significant amount of code from the Linear layer

### DIFF
--- a/src/nn/linear.rs
+++ b/src/nn/linear.rs
@@ -8,6 +8,7 @@ use std::borrow::Borrow;
 pub struct LinearConfig {
     pub ws_init: super::Init,
     pub bs_init: Option<super::Init>,
+    pub bias: bool,
 }
 
 impl Default for LinearConfig {
@@ -15,6 +16,7 @@ impl Default for LinearConfig {
         LinearConfig {
             ws_init: super::Init::KaimingUniform,
             bs_init: None,
+            bias: true,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


LaurentMazare:
Thanks for the PR. I like the idea of reusing of the existing linear layer in this case but maybe would it be possible to make it even simpler by adding the `bias` boolean in `LinearConfig` rather than having this `linear2` function.
As most use cases would customize `Default::default()` for the linear config, I would not expect this to break much code using this layer as long as we default to `true`.
The only drawback I can see is that it's possible to specify some `bs_init` value and to set `bias` to `false` which is a bit weird but I think it's reasonably fine (ideally we would have three cases for the bias: no-bias, bias-with-default-init, bias-with-custom-init, but switching to this is more likely to break existing code).
Let me know if this sounds reasonable. If so happy to get an updated PR for it or to make the changes myself if you prefer.

guillaume-be:
Makes sense - I was concerned of breaking existing implementations specifying the LinearConfig

PR #181, opened 2020-05-03
Add functions to load a JIT module directly on a specific device.

de-vri-es:
This PR adds functions to load a JIT module directly on a specified device.

This can be important, since it may otherwise be impossible to load a GPU module on a CPU when you don't have a CUDA enabled GPU. In that scenario, loading the GPU module already fails, so you can not move it to the CPU afterwards.

By exposing the device parameter of `torch::jit::load`, we can successfully load the model anyway :)

### DIFF
--- a/src/wrappers/jit.rs
+++ b/src/wrappers/jit.rs
@@ -260,12 +260,34 @@ impl CModule {
         Ok(CModule { c_module })
     }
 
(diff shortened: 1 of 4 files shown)


LaurentMazare:
Merged, thanks for the PR!

de-vri-es:
Thanks for the quick review :)

PR #183, opened 2020-05-03
Remove unnecessary braces around expressions.

de-vri-es:
While building locally, I got a lot of warnings about unnecessary braces around expressions. This PR removes them.

Some of the diffs are a bit larger because of reformatting by `cargo fmt`. If you prefer, I can undo the `cargo fmt` to make the diffs easier to read, and then you can run `cargo fmt` afterwards.

I'd remove the braces (hence this PR), but if you want to keep them it would make sense to explicitly silence the warning using a global `#![allow(unused_braces)]` at the crate root.

### DIFF
--- a/gen/gen.ml
+++ b/gen/gen.ml
@@ -523,9 +523,9 @@ let write_fallible_wrapper funcs filename =
           pm "    )%s {" (Func.rust_return_type func ~fallible:true);
           match func.returns with
           | `dynamic ->
-              pm "        let c_tensors = unsafe_torch_err!({";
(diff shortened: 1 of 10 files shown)


LaurentMazare:
I don't have this warning somehow, but that seems reasonable, thanks.

de-vri-es:
Probably due to different compiler version. My system defaults to nightly. To be specific, I'm now on 1.45.0-nightly from 2020-04-30. But now you'll never have to see those warnings ;)

Thanks for the library, and your general responsiveness. Much appreciated! :+1:

Issue #182, opened 2020-05-03
`Cannot initialize CUDA without ATen_cuda library`


johannesvollmer:
Hi! First, thanks for your work regarding PyTorch.

### Background
I have run into several problems when trying to run a project using [rust-bert](https://github.com/guillaume-be/rust-bert), a rust native Transformer-based models implementation which uses `tch-rs`. The CPU version ran just fine, but the CUDA version did not. Initially, I started [a thread on the rust-bert repository](https://github.com/guillaume-be/rust-bert/issues/24) with possibly more detailed information, but I'll summarize it here:

### Problem
First, switching from `Device::CPU` to `Device::CUDA` made it stop working and generated the following error: 

`TorchError { c_error: "Cannot initialize CUDA without ATen_cuda library. PyTorch splits its backend into two shared libraries: a CPU library and a CUDA library; this error has occurred because you are trying to use some CUDA functionality, but the CUDA library has not been loaded by the dynamic linker for some reason. The CUDA library MUST be loaded, EVEN IF you don\'t directly use any symbols from the CUDA library! One common culprit is a lack of -Wl,--no-as-needed in your link arguments; many dynamic linkers will delete dynamic library dependencies if you don\'t depend on any of their symbols. You can check if this has occurred by using ldd on your binary to see if there is a dependency on *_cuda.so library. (initCUDA at C:\\b\\windows\\pytorch\\aten\\src\\ATen/detail/CUDAHooksInterface.h:63)\n(no backtrace available)" }`.

Trying to fix it, I installed `CUDA 10.2.89`, updated graphics drivers, tried release and debug modes, always deleting the cargo build directory to force a fresh build. All of this did not change anything. 

Then I tried various manually installing `PyTorch 1.5`, setting environment variables (`LIBTORCH` and `PATH` with the LibTorch path, `TORCH_CUDA_VERSION` as `10.2`), but suddenly, the previous Error did not even show up, because a different runtime error aborted the process before anything else could happen: 

```error: process didn't exit successfully: `target\release\phrase-set-variations.exe` (exit code: 0xc0000135, STATUS_DLL_NOT_FOUND)```

__Now, not even the CPU version runs, not even after reverting the environment variable changes.__ :(
I was not able to resolve that error with Google, so I'm asking you for help here.

#### Environment
> `CUDA 10.2.89`
`tch = "0.1.7"`
`Windows 10`
`GeForce GTX 1060`
`rust-bert = "0.7.0"`

### Question
[...]

PR #184, opened 2020-05-03
Rerun build-script if important variables change.

de-vri-es:
This PR ensures that the build script for `torch-sys` is rerun when certain environment values are changed. This is limited to environment variables that change the behaviour of the build script that cargo doesn't already know about (so not the ones set by cargo).

This should prevent confusing situations like this: https://github.com/LaurentMazare/tch-rs/issues/182#issuecomment-623093543 . A clean build and or incremental build should normally be equivalent.

One downside of this PR: right now, if you build once with the right environment variables, subsequent runs of `cargo build` with the wrong environment variables won't break your build. On the other hand, this is somewhat unpredictable, as there might be other factors that lead cargo to rebuild your dependencies, and then the environment matters again.

All considered, I believe it's better that changing the environment for incremental builds gets picked up by the build script.

### DIFF
--- a/torch-sys/build.rs
+++ b/torch-sys/build.rs
@@ -66,10 +66,15 @@ fn extract<P: AsRef<Path>>(filename: P, outpath: P) -> Fallible<()> {
     Ok(())
 }
 
+fn env_var_rerun(name: &str) -> Result<String, env::VarError> {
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Looks good, thanks!

PR #185, opened 2020-05-05
Update rand to 0.7

danieldk:
A bit of a boring PR, but this deduplicates transitive dependencies (with the current version `rand_core` and `autocfg` are duplicates). It also reduces the number of transitive dependencies, since rand 0.7 depends on fewer rng implementations.


LaurentMazare:
Reducing dependencies sounds good, thanks for the PR.

PR #189, opened 2020-05-06
Introduce the TchError error type

danieldk:
Replaces the use of `failure` by library-specific errors.

The examples and tests use the `anyhow` crate, since their API is not
exposed.


LaurentMazare:
Thanks, this looks pretty good to me. I'll look at it a bit more today though.

LaurentMazare:
Merged, thanks again for the PR!

PR #192, opened 2020-05-19
Require Send on Module{,T} traits

jerry73204:
This patch adds additional trait bound `Module{,T}: Send` and changes `Func{,T}` and `Sequential{,T}` accordingly to satisfy the bound. It aims to ease the hassle to work with multi-threading.

It needs rigorous review on the `Func{,T}` type that it wraps around the `Box<dyn 'a + Fn(...) -> ... + Send>` type. It assumes the underlying closure or fn pointer could be moved among threads. I would be more specific that whether we can eliminate the case the closure can modify outer variables like `FnMut`.

I made a [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=82100a4074af0c93c3ea920ff2bd2f6f) to simulate the outcome of this patch. It verifies that closures (w/ or w/o capturing) and fn pointers compile.

### DIFF
--- a/src/nn/func.rs
+++ b/src/nn/func.rs
@@ -3,7 +3,7 @@ use crate::Tensor;
 
 /// A layer defined by a simple closure.
 pub struct Func<'a> {
-    f: Box<dyn Fn(&Tensor) -> Tensor + 'a>,
+    f: Box<dyn 'a + Fn(&Tensor) -> Tensor + Send>,
 }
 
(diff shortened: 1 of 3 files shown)


jerry73204:
I propose this change because when I use the efficientnet from `tch::vision` with asynchronous blocks, the lacking of `Send` blocks the interoperability with async/await.

LaurentMazare:
Merged, thanks for the PR.

PR #196, opened 2020-05-20
Internal tidy

jerry73204:
- Add more entries to `.gitignore`
- Make sure no public type is re-exported more than once. Those in `crate::wrapper::*` are re-exported in root, and any other children modules import from root directly. I make sure it does not change the public exported paths for compatibility.
- Split `tensor/mod.rs` into small modules for maintainability.

EDIT:
- Another reason not importing `crate::wrapper::*` types in children modules is that some types in `crate::wrapper::*` is dependent on code generation.
- Replace `TryInto` with `TryFrom`.


LaurentMazare:
I would like to push back a bit on this for now: if there are some specific bits around the gitignore that you want to be merged could you make a specific PR for these? As for the splitting of the tensor module I would have to think a bit more about it.

jerry73204:
The `.gitignore` patch is moved to #197.

jerry73204:
The library has gone thru great internal changes. The PR is no more relevant and we can consider closing it.

PR #198, opened 2020-05-20
Tensor conversion improvements

jerry73204:
- ~New `half::f16` half-precision floating point support. It can be enabled by additional `half-precision` feature.~ The `half::f16` will be moved to another PR.
- Reduce the usage of macros and use `tch::kind::T` trait instead.
- Add conversions from both owned and borrowed types `impl From<T> for Tensor` and `impl From<&T> for Tensor`
- New `Tensor::{from_shape_fn, from_shape}` to create a tensor from a callback or a vec/slice.
- Replace `From<_>` impls that could panic with `TryFrom<_>`.
- Now `impl `TryFrom<&Tensor> for Vec<_>` returns error if the tensor is not one-dimensional and fails on scalar tensors. Use `Tensor::to_flat_vec()` instead to get the flat vec.
- Unlike earlier versions, `TryFrom<&Tensor> for f64` does not implicitly cast from Float to Double.  Also, `TryFrom<&Tensor> for i64` fails on Bool tensor.
- Tests, examples and doc comments are changed accordingly to reflect the patch.
- Change the tests to assert equality of float tensors with `approx::assert_abs_diff_eq`.

~The PR is based on #196. It should be blocked until the changes of #196 is final.~ Now the PR is independent from #196.

I changed some tests and examples that are practically incorrect.
- The errors are mostly assertions on `assert_eq!(f32 , f32)`. They are rewritten to `approx::assert_abs_diff_eq`.
- Some tests implicitly convert a Float loss tensor to an `i64` integer. It no longer compiles in the PR due to strict conversion rule, and now it's changed to floats and tested by `assert_abs_diff_eq`.
- The libtorch use Float and Int64 by default, which are distinct from Rust's default literal type `f64` and `i32`. I changed many `f64::from(tensor)` to `f32::try_from(tensor)` in many tests and examples. I made my best effort to test examples to avoid runtime errors on wrong types. However, I would appreciate some one else can verify it.


jerry73204:
I realized the refactoring work is non-trivial. I'm WIP on re-writing the conversion again based on master branch. There are several design decision to make and I need others' options to make it clear.

- Does `From<Tensor> for Vec<_>` assume the tensor is one-dimensional? If yes, times to refactor tests. If not, it's inconsistent with the semantic of `From<Tensor> for Vec<Vec<...>>` and it's better to move to `Tensor::to_shape_vec()`.
- Some conversions, like`i64::from(non_scalar_tensor)`, could panic. Suppose we implement `TryFrom` instead? If yes, the refactoring is going. If not, how could we provide an alternative interface that does not kill my program? Note that `From<_>` and `TryFrom<_>` will conflict with each other.
- `i64::from(tensor)`  implicitly converts a non-i64 tensor to an integer. It incurs undesired side-effect if the precision of tensor is critical (such as loss). I observed such errors in several tests. Perhaps we can make is strict.

Apart from issues above, some improvements are already done.
- Given we have `tch::kind::T`, some conversions can be defined using generics instead of macros. `From<Tensor> for $primitive` can be done this way.
- The conversion of `ndarray::Array` can be more generic, and is not restricted to `ArrayN`.

jerry73204:
I pushed new commits and the CI passes. It's ready for review.

LaurentMazare:
Sorry this PR tweaked lots of things in various ways so I never got around reviewing it. If you still have bits that you feel would be useful, please create an issue first with a scope as limited as possible so that it can be discussed there first.

PR #199, opened 2020-05-22
Fix rl

spebern:
Hi,

The current rl examples did not compile because the optimizer needs to be mutable.
Also, there was a runtime error in the pg example that I took care of.

Thank you for this wrapper!

All the best,
Ben


LaurentMazare:
Thanks for the PR! This bit is not compiled in our CI test which explains why it was broken, maybe we should tweak the CI at some point to also compile these examples.

PR #200, opened 2020-05-24
Add ddpg example

spebern:
Hi,

I've added an example for [ddpg](https://arxiv.org/abs/1509.02971).

I had to make a couple of changes so that it works with the pendulum environment which can probably be improved and I am very open to suggestions on that.

I hope that this is helpful to someone!

All the best,
Ben


LaurentMazare:
That's a very nice addition, thanks for the PR. The code looks quite good to me (and the changes to `Step` are very reasonable).

Issue #100, opened 2019-08-20
Type mismatch error with Linear layer


jean-airoldie:
I get an errror: `Expected object of scalar type Double but got scalar type Float for argument`
caused by:
https://github.com/LaurentMazare/tch-rs/blob/c6888ba746d7123939bf8f092c06b35a16319b7c/src/nn/linear.rs#L51
when passing a `Kind::Double` to the `forward` method of a `Linear` layer. Seems like there is a type mismatch between the weights and biases.

You can find the full backtrace [here](https://gist.github.com/jean-airoldie/69c4a9c2e5c717bd59dc4dd2b6fe47ec).

Here is the minimal reproducible code:
```rust
use tch::{nn, nn::Module, nn::OptimizerConfig, Device, Tensor, Kind, Reduction};

const INPUT_DIM: i64 = 784;
const BATCH_SIZE: i64 = 1;
const HIDDEN_NODES: i64 = 128;
const LABELS: i64 = 10;

fn net(vs: &nn::Path) -> impl Module {
    nn::seq()
        .add(nn::linear(
            vs / "layer1",
            INPUT_DIM,
            HIDDEN_NODES,
            Default::default(),
        ))
        .add(nn::linear(vs, HIDDEN_NODES, LABELS, Default::default()))
}

pub fn run() -> failure::Fallible<()> {
    let vs = nn::VarStore::new(Device::Cpu);
    let net = net(&vs.root());
    let opt = nn::Adam::default().build(&vs, 1e-3)?;
    for epoch in 1..200 {
        let features = Tensor::rand(&[BATCH_SIZE, INPUT_DIM], (Kind::Double, Device::Cpu));
        let labels = Tensor::rand(&[BATCH_SIZE, LABELS], (Kind::Double, Device::Cpu));

        let predictions = net.forward(&features); // This panics.

        let loss = predictions.mse_loss(&labels, Reduction::Mean);
        opt.backward_step(&loss);
        println!(
            "epoch: {:4} train loss: {:8.5}",
            epoch,
            f64::from(&loss),
        );
    }
    Ok(())
}
```

LaurentMazare:
Yes that's actually expected: all the layers in `nn` use single precision float and not double precision ones. It feels like the typical use case for neural networks - although supporting half-precision float would be nice.
I imagine you already know about this but a quick workaround for your use case is to use `Kind::Float` for `features` and `labels`.
[...]

PR #204, opened 2020-06-03
Support tensor ops with i32 and f32 types along with improvements

jerry73204:
Changes:
- Define {Add,Sub,Mul,Div}{,Assign} on i32 and f32 with tensor
- Replace macros with generic impl
- Reuse existing code and remove duplicated parts

It could somewhat solves #100. However, with this PR, the `1 + tensor` should be more verbose `1i64 + tensor`, otherwise it won't compile.

### DIFF
--- a/examples/vae/main.rs
+++ b/examples/vae/main.rs
@@ -57,7 +57,7 @@ fn loss(recon_x: &Tensor, x: &Tensor, mu: &Tensor, logvar: &Tensor) -> Tensor {
     //     Kingma and Welling. Auto-Encoding Variational Bayes. ICLR, 2014
     // https://arxiv.org/abs/1312.6114
     // 0.5 * sum(1 + log(sigma^2) - mu^2 - sigma^2)
-    let kld = -0.5 * (1 + logvar - mu.pow(2) - logvar.exp()).sum(Kind::Float);
+    let kld = -0.5 * (1i64 + logvar - mu.pow(2) - logvar.exp()).sum(Kind::Float);
     bce + kld
 }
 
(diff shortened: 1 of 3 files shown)


LaurentMazare:
The i32/f32 support seems nice.
What is the advantage of the generic impl, this seems to introduce a bunch of boilerplate that wasn't needed before, is it necessary to get the i32/f32 bit?

jerry73204:
Generic blocks and `i32/f32` are two things. You can add types in in either macros or generic blocks by your choice. Replacing macros with generics is an additional change along with `i32/f32`. Here are my reasons.

Think of adding a new type to `Scalar`. We have to add macro calls by using macro_rules, while generic impls will automatically build one for you. In user's perspective, the error message in plain impl would be more approachable than in macros.

Another minor reason is to avoid possible performance pitfall on macro expansions, though ours is not the case. The [reddit post](https://www.reddit.com/r/rust/comments/er1xkh/how_refactoring_a_macro_reduced_my_rust_project/)  is a great example for those interested.

The macro expands impl blocks for each of `f64` and `i64` before this PR. By replacing some with generics, the code length is actually shortened. The code length in this PR looks the same because I add more impls for owned `Tensor` in the macro. It's yet another improvement.

Maybe future PRs would add `f16` and complex number types. This change will make it less painful.

jerry73204:
If you feel like the original writing is more superior, we can move on to find the best solution.

LaurentMazare:
I feel that the new implementation is a bit more verbose and I'm not sure there is much to gain on the performance of macro expansion - the current compilation time is mostly limited by compiling the gigantic rust and c++ files that get generated - compiling this file is probably negligible.
Anyway I don't have a super strong opinion neither so I've merged the change, sorry for the delay.

PR #214, opened 2020-06-13
Improve EfficientNet code quality

jerry73204:
The changes are aimed for better user experiences.

- Give explicit return types instead of `impl ModuleT`
- Use generic arguments
- Make parameter struct and initializing methods public
- Write doc comments

### DIFF
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -629,6 +629,14 @@ impl Tensor {
         result.copy_(&self);
         result
     }
+
+    pub fn f_swish(&self) -> Result<Tensor, TchError> {
+        Ok(self * self.f_sigmoid()?)
+    }
+
+    pub fn swish(&self) -> Tensor {
+        self.f_swish().unwrap()
+    }
 }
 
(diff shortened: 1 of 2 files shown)


jerry73204:
I come up a question. What's the point to give `Func<'a>` an explicit life-time ([code](https://github.com/LaurentMazare/tch-rs/blob/488d7e2d71395eb864c8296a21f865e989c70b3f/src/nn/func.rs#L5-L7)). IMO, 90% of the time we can get around with `'a = 'static` for function pointers and closures. I wonder if any rare cases that `'a` cannot be `'static`.

```rust
struct Func<'a>(pub Box<dyn 'a + Fn() -> ()>);

fn foo() {
    println!("world");
}

fn main() {
    let bar = || {
        println!("hello");
    };

    let _: Func<'static> = Func(Box::new(foo));
    let _: Func<'static> = Func(Box::new(foo.clone()));
    let _: Func<'static> = Func(Box::new(bar));
}
```

LaurentMazare:

PR #221, opened 2020-06-24
Added support for f16 serialization

guillaume-be:
Following the support for f16 tensors in https://github.com/LaurentMazare/tch-rs/pull/215, could we please add the support to load/read numpy arrays with half precision?

I just came across an example and the current setup requires me to serialize the weights with f32 precision and therefore ~ double the file size compared to the native Pytorch parameters.


LaurentMazare:
Looks all good, thanks for the PR!

PR #224, opened 2020-06-30
Add WGAN-GP example

jerry73204:
I'd like to present the Wasserstein GAN with gradient penalty example. It is slightly modified from the [gan example](https://github.com/LaurentMazare/tch-rs/blob/master/examples/gan/main.rs). It removes batch norms in discriminator as the paper suggested, and replace them with instance norms.

I tested against the MNIST images. It works fine on single channel (grayscale) image.
https://imgur.com/k5nQryZ

However, I found the image is bulrring if three-channel RGB images are provided.
https://imgur.com/GuCVnEy

The code is almost ready. If anyone can tell if it is an implementation or fine-tuning issue, I could fix it soon and merge.

### DIFF
--- a/examples/wgan/main.rs
+++ b/examples/wgan/main.rs
@@ -1,0 +1,364 @@
+use anyhow::{bail, Result};
+use std::{borrow::Borrow, path::Path};
+use tch::{
+    kind::{Kind, INT64_CPU},
+    nn::{self, OptimizerConfig},
+    Device, Tensor,
+};
+
+// model parameters
+const IMAGE_DIM: i64 = 3;
+const LATENT_DIM: i64 = 128;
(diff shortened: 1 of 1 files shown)


jerry73204:
I tested WGAN-GP and it also generated blurry images. I found sources that experienced with this issue ([1](https://github.com/AlexiaJM/Deep-learning-with-cats/issues/15), [2](https://openreview.net/pdf?id=SJx9GQb0-)).

I see we already have a Realtivistic DCGAN example. WGAN is just a plus. I not sure about the plan. Perhaps we can
- Discard this PR
- Accept this PR but without RGB blurry example
- Accept this PR with RGB blurry example.

LaurentMazare:

PR #233, opened 2020-07-20
Define separate fallible f_sizeN methods with N up to 6

jerry73204:
It follows the convensions to separate `sizeN` and `f_sizeN` methods, and extends to N up to 6. I found higher dimension count is needed when writing YOLO, for example `[batch_size, anchors, height, width, outputs]`.

### DIFF
--- a/examples/neural-style-transfer/README.md
+++ b/examples/neural-style-transfer/README.md
@@ -55,7 +55,7 @@ define a loss that is location independent.
 
 ```rust
 fn gram_matrix(m: &Tensor) -> Tensor {
-    let (a, b, c, d) = m.size4().unwrap();
+    let (a, b, c, d) = m.size4();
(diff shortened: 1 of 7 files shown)


LaurentMazare:
Thanks for the PR.
I don't think we would need to follow the `f_` idiom for size. This would break existing users and I'm not sure it provides much: the main reason to have panicking function is to make the "formulas" part in the code more readable. However here we would de-structure the tuple on the same line so having either an `unwrap()` or a `?` doesn't sound too bad and is more explicit.
`size5` and `size6` indeed seem interesting to have, I've imported that part of the change.

jerry73204:
> the main reason to have panicking function is to make the "formulas" part in the code more readable

Sounds a good argument here. I wish I knew this statement somewhere else when contributing to the project. Thanks for your change.

PR #237, opened 2020-08-03
Add the autocast function to run a closure in mixed-precision

danieldk:
I have made this a draft PR, because I can't currently test this (all our GPUs that support mixed-precision have jobs running for multiple days). Also, I haven't added any unit tests, since actually testing this would require a CUDA-capable GPU.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ members = ["torch-sys"]
 [features]
 python = ["cpython"]
 doc-only = ["torch-sys/doc-only"]
+autocast-tests = []
 
 [package.metadata.docs.rs]
 features = [ "doc-only" ]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,8 +21,8 @@ pub use wrappers::{
 
 mod tensor;
 pub use tensor::{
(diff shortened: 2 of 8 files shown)


LaurentMazare:
Looks very nice thanks for the PR. Not having unit tests run automatically sounds reasonable but maybe we could have some tests or an example behind some feature gate ?
Also note that the appveyor CI fails because of missing autocast symbols on the windows build. I haven't investigated but maybe these symbols are not available on windows somehow?

danieldk:
> Looks very nice thanks for the PR. Not having unit tests run automatically sounds reasonable but maybe we could have some tests or an example behind some feature gate ?

I'll try to add some tests behind a feature gate.

> Also note that the appveyor CI fails because of missing autocast symbols on the windows build. I haven't investigated but maybe these symbols are not available on windows somehow?

I think this was caused by name mangling. The new symbols were not in `torch_api.h`, so not within `extern "C"`. Fixed now.

danieldk:
I have added an `autocast-tests` feature, which runs a simple test case (with nesting of `autocast`). I passes with flying colors on a Tesla P100.

LaurentMazare:
Merged, thanks for the PR!

PR #245, opened 2020-09-04
Add Tensor::is_mkldnn

danieldk:
This makes it possible to implement model conversion similar to:

https://github.com/pytorch/pytorch/blob/master/torch/utils/mkldnn.py


LaurentMazare:

PR #246, opened 2020-09-04
Add the ToMKLDNN trait and implement it for Linear

danieldk:
I am not sure if this functionality/interface fits the tch crate, but this attempts to make something close to

https://github.com/pytorch/pytorch/blob/master/torch/utils/mkldnn.py

Some people have reported large improvements with something as simple as converting all BERT linear layers to use MKL-DNN:

https://github.com/pytorch/pytorch/pull/21851

It's a bit annoying that these things are not automatically done, but MKL-DNN/oneDNN require different ordering of dimensions.

If this doesn't fit the `tch` crate, no big loss, I can just import it in my own projects ;).

### DIFF
--- a/src/nn/linear.rs
+++ b/src/nn/linear.rs
@@ -57,6 +57,30 @@ pub fn linear<'a, T: Borrow<super::Path<'a>>>(
 
 impl super::module::Module for Linear {
     fn forward(&self, xs: &Tensor) -> Tensor {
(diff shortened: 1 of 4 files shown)


danieldk:
I have tested this a bit in one of my own projects, and it is quite invasive. It is probably better to use mkldnn tensors and operations locally in a focused manner. So, I'll close this PR for now.

PR #249, opened 2020-09-12
Add Tensor::internal_amp_non_finite_check_and_unscale

danieldk:
This function unscales a tensor and at the same time checks for
infinities.

This is one of the methods needed to implement gradient scaling for
auto mixed precision.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ members = ["torch-sys"]
 [features]
 python = ["cpython"]
 doc-only = ["torch-sys/doc-only"]
-autocast-tests = []
+cuda-tests = []
 
 [package.metadata.docs.rs]
 features = [ "doc-only" ]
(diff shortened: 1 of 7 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks for the PR!

danieldk:
> Thanks for the PR!

Thanks for merging! All the primitives for auto mixed-precision training training should now be in place. I guess it would be nice to have a more user-friendly abstraction for gradient scaling, like PytTorch' `GradScaler`. I something working in one of my projects (still work in progress):

https://github.com/stickeritis/sticker2/blob/mixed-precision/sticker2/src/optimizers/grad_scale.rs

But I don't use the `tch:nn` optimizers, because I needed AdamW with discriminative learning rates/fine tuning and I assume that only one device/GPU is used. But I could look into generalizing it for doing a PR later (unless someone else beats me to it ;)).

PR #257, opened 2020-10-03
Add support for parameter groups

danieldk:
I have been working on support for discriminative learning rates in
optimizers, as e.g. PyTorch supports. Before polishing a bit more and
perhaps adding tests, I wanted to ask if you would be interested in
supporting discriminative learning rates, and if the proposed API
would be ok for you.

The commit message follows ;).

---

The tch crate places all parameters in the same group. This simplifies
construction of optimizers, however it does not allow training models
with discriminative learning rates as is e.g. common in finetuning
pretrained transformers such as BERT.

This change adds support for training with discriminative learning
rates. In order to do so, `OptimizerConfig` provides an extra
constructor, `build_with_groups`. Compared to the existing `build`
contructor, `build_with_groups` allows the caller to provide a
closure that partitions variables (by name) into parameter groups.
For example:

```rust
adam.build_with_groups(&vs, lr, |var_name| {
  if var_name.starts_with("classifier") {
    PARAMETER_GROUP_CLASSIFIER
  } else if var_name.starts_with("encoder") {
    PARAMETER_GROUP_ENCODER
  } else { unimplemented!() }
});
```

This can then later be used to set the learning rate per group.  For
example:

```rust
optimizer.set_lr_group(PARAMETER_GROUP_CLASSIFIER, classifier_lr);
```

In order to support group partitioning, `trainable_parameters` of
`VarStore` had to be changed into a `HashMap`, so that the names of
trainable parameters are known.

### DIFF
--- a/examples/custom-optimizer/sparse_adam.rs
+++ b/examples/custom-optimizer/sparse_adam.rs
@@ -58,7 +58,7 @@ impl SparseAdam {
             .lock()
             .unwrap()
             .trainable_variables
-            .iter()
+            .values()
             .map(|x| Buffer::new(&x.size()))
             .collect();
 
@@ -82,7 +82,7 @@ impl SparseAdam {
         let mut vars = self.vars.lock().unwrap();
 
         // iterate through all trainable variables
-        for (tensor, buffer) in vars.trainable_variables.iter_mut().zip(&mut self.buffers) {
+        for (tensor, buffer) in vars.trainable_variables.values_mut().zip(&mut self.buffers) {
             let mut grad = tensor.grad();
 
             // calculate both bias correction values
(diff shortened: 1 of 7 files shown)


danieldk:
Not sure why Windows CI fails, the tests run fine on my Linux machine.

LaurentMazare:
Thanks for working on this and sorry for taking this long to come back to this. Supporting params group is certainly worthwhile (multiple varstore can probably be used for different learning rate but that sounds less convenient when serializing/loading models).

Looking at your PR, I feel that matching on the variable name might be a bit error-prone so I've crafted a slightly different change #257 It's fairly similar in that it provides a way to set learning-rate/momentum for some group, e.g. via `opt.set_lr_group(GROUP, lr)`. However changing the group is done in a similar way the variable path gets modified, e.g.:

```rust
let root = vs.root();
let var = root.set_group(1).zeros("varname", &[]); // Create a variable with group 1
let modl = create_module(&root.set_group(5)); // Use group 5 when calling create_module
```
  
The default group is 0, and `set_lr` can still be used to tweak all the param groups at once.

If you still want to base the group on variable names, you can probably in your model warp `VarStore` or `Path` to include the closure that maps names to groups.

danieldk:
Sounds good!

LaurentMazare:
I've merged the changes, and will have a look at tweaking the transformer example if relevant. Thanks again for all the ground work on this!

PR #258, opened 2020-10-07
Add examples on tensor index/put

evgerher:
Could not find previously needed examples of tensor index/put operations in provided tests/examples, thus created own tests for demonstration purposes.


LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR!

PR #266, opened 2020-10-17
Add Path::components, which returns an iterator over the path components

danieldk:
I have been implementing an extension to `Path` to assign groups by
name as proposed in #257. However, in order to do so, I need access to
the path components.

This function returns an iterator over the components (similar to
`std::path::Path`), though returning `&[String]` is also an option.
Though returning an iterator should make it easier to change internals
without changing the API in the future.

### DIFF
--- a/src/nn/var_store.rs
+++ b/src/nn/var_store.rs
@@ -221,6 +221,11 @@ impl VarStore {
 }
 
 impl<'a> Path<'a> {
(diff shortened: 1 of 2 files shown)


LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR.

PR #267, opened 2020-10-17
Remove unused `config` field of `Optimizer`

danieldk:
This makes `Optimizer` easier to use downstream, since it does not
carry a type parameter anymore.


LaurentMazare:
This config field is actually useful in some of my use cases, so I would prefer keeping it.

danieldk:
Ack.

PR #269, opened 2020-10-18
Add Optimizer::set_weight_decay(_group)?

danieldk:
This makes it possible to set the weight decay per group, which is
useful to exclude e.g. layer normalization from weight decay.


LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR!

Issue #265, opened 2020-10-16
Loading a tensor from device data pointer.


omerbenamram:
Hi there!

I'm trying to create a `tch::Tensor` from existing inference output data on a CUDA device.
I was trying to see if there is the API described in [https://discuss.pytorch.org/t/create-tensor-from-data-pointer/19708](https://discuss.pytorch.org/t/create-tensor-from-data-pointer/19708) exists, but `tch` doesn't seem to have the notion of context objects, not the `fromBlob` api.

If I have a data pointer on the host, this indeed seems to work:
```
        Tensor::of_data_size(
            std::slice::from_raw_parts(
                self.data_ptr,
                (self.number_of_elements * 2) as usize, // *2 because of fp16
            ),
            &self.dims,
            tch::Kind::Half,
        )
```

But I haven't found any API that also accepts this on a CUDA device.

Is this something that makes sense considering the architecture of `tch`?

Thanks alot for this wonderful crate :)

omerbenamram:
I did end up implementing the wrapper around `torch::from_blob`, and this works, however as per the docs, `tch` will raise an exception if we try to manipulate the memory afterwards, but it did solve my case of a read only view into an existing CUDA mapped data.

https://github.com/LaurentMazare/tch-rs/compare/master...omerbenamram:feature/tensor_of_blob

omerbenamram:
closed by #273.

PR #273, opened 2020-10-27
Support externally created tensors

omerbenamram:
Following #265, I was wondering if this API addition is something consider accepting into upstream.

It resulted in massive performance wins for us when manipulating huge tensors that are externally allocated by TensorRT.
Are there any changes that should be added?

### DIFF
--- a/src/wrappers/tensor.rs
+++ b/src/wrappers/tensor.rs
@@ -451,7 +451,7 @@ impl Tensor {
     /// Creates a tensor from data that is assumed to be initialized.
     /// Resize operations are now allowed on this tensor without copying the data first.
     /// # Safety
-    ///   This will panic if `data` is points to invalid data.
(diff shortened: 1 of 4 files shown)


src/wrappers/tensor.rs
```diff
@@ -448,6 +448,40 @@ impl Tensor {
         Ok(Tensor { c_tensor })
     }
 
+    /// Creates a tensor from data that is assumed to be initialized.
+    /// Resize operations are now allowed on this tensor without copying the data first.
+    pub fn f_of_blob(
```
danieldk, line 453: This should also be unsafe, right? Since data can be an invalid pointer.

omerbenamram: yes, I've changed this to be unsafe as well, and added clippy SAFETY docs.

LaurentMazare:
Merged, thanks for the PR!

PR #289, opened 2020-11-24
Add quantized  JIT example

calmdown13:
## Summary
Small adaptation of the JIT example to demonstrate how to load quantized models.

### DIFF
--- a/examples/jit-quantized/main.rs
+++ b/examples/jit-quantized/main.rs
@@ -1,0 +1,54 @@
+// This example illustrates how to use a quantized PyTorch model trained and
+// exported using the Python JIT API. Add
+// See https://pytorch.org/tutorials/advanced/cpp_export.html and
+// https://pytorch.org/docs/stable/quantization.html for more details.
+use std::time::SystemTime;
+extern crate tch;
+use anyhow::{bail, Result};
+use tch::vision::imagenet;
+
+pub fn main() -> Result<()> {
+    let args: Vec<_> = std::env::args().collect();
+    let (model_file, image_file, qengine) = match args.as_slice() {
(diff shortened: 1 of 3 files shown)


examples/jit-quantized/main.rs
```diff
@@ -0,0 +1,54 @@
...
+    // Load the image file and resize it to the usual imagenet dimension of 224x224.
+    let image = imagenet::load_image_and_resize224(image_file)?;
+
+    // Load the Python saved module.
+    let model = tch::CModule::load(model_file)?;
+
+    // Time inference
+    let now = SystemTime::now();
+    for _ in 1..10 {
+        image.unsqueeze(0).apply(&model);
```
LaurentMazare, line 35: Could you fix the warning there about the output being unused, e.g. adding a `let _output =` at the beginning of this line?

examples/jit-quantized/resnet.py
```diff
@@ -0,0 +1,22 @@
+import torch
+import torchvision
+
+
+def save_resnet18_script_module(
+    filename: str, pretrained: True, quantize: False
```
LaurentMazare, line 6: I'm not very familiar with python type annotations, but do True/False make sense in this setting or should these be `bool` (or default values, e.g. `pretrained=True`.

examples/jit-quantized/main.rs
calmdown13: sure, updated

examples/jit-quantized/resnet.py
calmdown13: woops my bad, updated

LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR!

PR #294, opened 2020-11-29
Support loading of i8 npz files generated by numpy

guillaume-be:
Hello,

Numpy's description for the `np.int8` type is `|i1`, which unfortunately doesn't match the pattern used in the current library when loading `npz` files:
https://github.com/LaurentMazare/tch-rs/blob/660a20de43898ed473c2a1a41a355ec300007064/src/tensor/npy.rs#L151

This PR proposes adding support for loading these int8 numpy arrays, useful for quantized models

### DIFF
--- a/src/tensor/npy.rs
+++ b/src/tensor/npy.rs
@@ -148,7 +148,7 @@ impl Header {
                         descr
                     )));
                 }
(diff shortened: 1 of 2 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks for catching this and for the PR!

Issue #281, opened 2020-11-17
Discussion on Tensor type conversions


jerry73204:
It is due to my closed #198. It is outdated for a while and I think it should be discarded too. Nevertheless, let me leave some concerns about type conversions in current implementation. Some changes would be made as soon as some concerns become clear.

- Implicit floating point truncation
It is able to divide a `Float` kind tensor with a `f64` number. It involves implicit f64 to f32 truncation. We might restrict the divisor to be `f32`, but it sacrifices the convenience by writing `tensor / 2f32` everywhere. I consider a balanced approach is that we allow implicit conversions for `+-*/`, and provide explicit method calls that restricts scalar types.

- Deprecate `From<Vec<Vec<T>>> for Tensor`
The [std doc](https://doc.rust-lang.org/std/convert/trait.From.html) says `From` must not fail. Also, it is discouraged to build vec of vec to store higher dimensional data because it is prone to error and is less efficient, though this feature is convenient for hacking. We could follow ndarray's [`from_shape_vec()`](https://docs.rs/ndarray/0.13.1/ndarray/struct.ArrayBase.html#method.from_shape_vec) or provide a fallible `from_guessed_vecs()` that make shape guessing explicit.

- Convenient `Tensor` constructors
I wish to have `from_shape_fn` and `from_shape_iter` like ndarray. It really saves my time.

- Third-party type conversion
I wish it works with well-known crates without much effort. It's nice we already have `ndarray`. It's better it supports `nalgebra` and `image`. I see introducing them would cause bloating. It could be solved by cargo feature gating or introduce an extra crate for type conversion.

- Let `Tensor` to be `Copy` (and of course `Clone`)
I see it is a controversial decision. Let me argue that it is valid that `Tensor` can be `Copy`. My retrospection on `Clone` and `Copy` is that it affects `.clone()` is explicit or not. Costly copy should have explicit `.clone()`, while `Copy` is suitable for primitive or cheap reference types. IMO, libtorch's `Tensor` is a reference to shaped data on device (so we have `.shallow_clone()`). We could follow PyTorch's approach to have implicit shallow copy and explicit deep `.copy()` method.
By having `Copy` on `Tensor`, we gain some conveniences. We can keep `Add<Tensor> for Tensor` and remove `Add<&Tensor> for Tensor` without bothering the borrow checker. It's also less cumbersome to derive `Clone` on struct of `Tensor`s.

edlanglois:
[...]

PR #297, opened 2020-12-07
Revise TryFrom impls and make ndarray optional

jerry73204:
`TryFrom` now receives a referenced type instead of owned type. The move is unnecessary because it always copies the data.

`TryInto` impls are removed in flavor of `TryFrom`. That's because `TryFrom` implies `TryInto`, but it does not in vice versa.

The ndarray crate becomes an optional but enabled by default feature to avoid bloating. It helps reducing compiling time for those who does not need this feature.


jerry73204:
BTW, I noticed the inconsistency of `TryFrom<Vec<T>>` and `From<&[T]>`. The latter does not follow the guideline that `From<_>` should not fail.

Also. regarding to #281, we have lots of `From<T> for Tensor` that could panic. However, moving to `TryFrom` is a breaking change. How do you think @LaurentMazare?

jerry73204:
I rebase the patch and upgrade ndarray to 0.14. To recap, the PR made following changes:

- Replace `TryInto` to `TryFrom` as suggested by std doc
- `ndarray` becomes an optional but enabled by default cargo feature, to reduce bloating
- Update to 0.14

The only breaking change is ndarray version update to 0.14.

PR #310, opened 2021-01-14
Fix seq2seq translation example

VasanthakumarV:
This PR makes the following changes to make the translation example run,
* Removes the sequence dimension from encoder output (this is required when we stack up the encoder outputs)
* Removes the sequence dimension from decoder output (this is required when the model output is compared with the target)
* Removes the sequence dimension from the hidden state vector during a concatenation operation in the decoder


LaurentMazare (approved):

LaurentMazare:
Merged, thanks!

PR #314, opened 2021-01-18
Update dependencies via cargo-edit

jerry73204:
This PR upgrades dependencies using `cargo upgrade` from [cargo-edit](https://crates.io/crates/cargo-edit).
Note that it introduces breaking changes due to `cpython` and `half` update, requiring minor version bump.


LaurentMazare:

PR #318, opened 2021-01-22
Implement binding generator in Rust

jerry73204:
The PR totally rewites the OCaml binding generator in Rust. Now the re-generated code passes every test on tch-rs and on large projects including [rust-bert](https://github.com/guillaume-be/rust-bert) and my [yolo-dl](https://github.com/jerry73204/yolo-dl).

My motivation was to add libtorchvision bindings, but stuck in the OCaml bindgen [code](https://github.com/LaurentMazare/tch-rs/tree/master/gen). I felt the pain in the hassle of codegen in strings. The rewriting will gain us the Rust's safety feature and make the project self-hosted. (I did not blame OCaml. It's a good language and I learned it recently.)

- Declaration file deserialization is done with [serde](https://github.com/serde-rs/serde) in declarative way.
- Generated Rust code is assembled with [quote](https://github.com/dtolnay/quote) token by token rather in strings.
- It becomes easy to refactor and maintain. Every stage is explicitly and statically typed. The compiler type checking will be a good aid.

The new generated bindings should maintain the compatibility with previous code. It is ensured by tests some manual audit on rustdoc. To check if the new bindings works for your project, run `make gen` in tch-rs directory and add this to your project `Cargo.toml`.

```rust
[patch.crates-io]
tch = { path = "/path/to/tch-rs" }
torch-sys = { path = "/path/to/tch-rs/torch-sys" }
```

BTW, I'm working on libtorchvision bindings. It will be easier if this PR can be accepted.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ cpython = { version = "0.2.0", optional = true }
 anyhow = "1"
 
 [workspace]
-members = ["torch-sys"]
+members = ["torch-sys", "tch-bindgen"]
 
 [features]
 python = ["cpython"]
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ clean: .FORCE
 	cargo clean
 
 gen: .FORCE
-	dune exec gen/gen.exe
+	cargo run --bin tch-bindgen --manifest-path tch-bindgen/Cargo.toml --release
 	rustfmt src/wrappers/tensor_fallible_generated.rs
 	rustfmt src/wrappers/tensor_generated.rs
 	rustfmt torch-sys/src/c_generated.rs
--- a/tch-bindgen/Cargo.toml
+++ b/tch-bindgen/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "tch-bindgen"
+version = "0.1.0"
+authors = ["jerry73204 <jerry73204@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.38"
+fstrings = "0.2.3"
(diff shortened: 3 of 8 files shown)


doofin:
Looks great! hasktorch also generate bindings from Declarations.yaml,what is left to be done?

doofin:
Just played around with this PR,build without problem and MNIST example works(on newest arch linux)

jerry73204:
> Looks great! hasktorch also generate bindings from Declarations.yaml.Does this use rust-bindgen?

@doofin It generates every function signature according to `Declarations.yaml` without `bindgen`.

>  BTW,does this imply that we will get the same api as the pytorch cpp api?

Basically yes, `Declarations.yaml` is generated by compiling libtorch.

LaurentMazare:
It is already the case that the Rust api in this create is automatically generated from `Declarations.yaml` and it has been so since the first version of this crate. You can find the generated api in [tensor_generated.rs](https://github.com/LaurentMazare/tch-rs/blob/master/src/wrappers/tensor_generated.rs) for example.
The design goal was indeed to keep the api as close as possible to the C++ one (which itself is pretty close  to the Python version), so that it's easier to port examples and applies existing tutorials. And it should be possible to write more Rust idiomatic crates on top of this one.
The part generating the code from `Declarations.yaml` is actually in OCaml as it's shared with the [ocaml-torch](https://github.com/LaurentMazare/ocaml-torch) sister project. OCaml is not required to compile this crate though as the generated files are checked in in the repo. This PR rewrites the code generation in Rust, however we would like to keep it in OCaml at the moment so as to continue sharing code and functionalities with ocaml-torch.

doofin:
Great,maybe it's better to start a new separate project for codegen.

jerry73204:
Seems it is decided to couple the OCaml project, though I think the opposite. I will move the code to another repo for those interested.

Issue #311, opened 2021-01-14
Support Exporting to TorchScript


tiberiusferreira:
Hello, thanks for all the time and effort put in this project.

I thought that exporting TorchScript from C++ (and by extension, Rust) was not possible due to https://discuss.pytorch.org/t/how-to-save-torchscript-model-using-cpp/59743.

However, the Haskell binding appear to support it: https://github.com/hasktorch/hasktorch/issues/302

Exporting to TorchScript is important to be able to emit ONNX ([which can be generated from TorchScript](https://pytorch.org/docs/stable/onnx.html)) models which then can be optimized and run in another runtime, such as [TensorRT](https://developer.nvidia.com/Tensorrt), [ONNX Runtime](https://github.com/microsoft/onnxruntime/tree/master/orttraining/orttraining) or used in iOS/Android or any other environment without shipping the entire PyTorch runtime.

Is it something this project would like to support?

LaurentMazare:
Hello,
Thanks for pointing this out, I didn't know that this existed on the C++ side, this seems pretty neat. I'll have a look at whether this is easy to wrap in this crate.

LaurentMazare:
Closing this now that `tch::CModule::create_by_tracing` is available, feel free to re-open if you encounter any issue/missing bits.

JinXiaozhao:
@LaurentMazare, I have no idea when create_by_tracing  should be used. Could you give an example?

LaurentMazare:
There is an example using this in the [jit tests](https://github.com/LaurentMazare/tch-rs/blob/45a0ac884e61d7305e4c6e2fae9497504000fe31/tests/jit_tests.rs#L131).

PR #319, opened 2021-01-23
Add support for the tracer api

LaurentMazare:
As suggested in #311, this adds support for the tracer api.
The implementation however is quite rough:
- The number of outputs has to be specified manually (and this is unsafe as an error here might lead to a segfault).
- Errors raised within the closure are not handled properly.
- There might remain some memory leaks too.

This snippet extracted from the test shows how this can be used.
```rust
    let closure = |inputs: Vec<Tensor>| {
        let v1 = inputs[0].shallow_clone();
        let v2 = inputs[1].shallow_clone();
        vec![v1 + v2]
    };
    let modl = tch::CModule::create_by_tracing(
        "MyModule",
        "MyFn",
        &[Tensor::from(0.0), Tensor::from(1.0)],
        1,
        &closure,
    )
    .unwrap();

    // Save the module in a file.
    let filename = std::env::temp_dir().join(format!("tch-modl-{}", std::process::id()));
    modl.save(&filename).unwrap();

    // Reload the module to try it out.
    let modl = tch::CModule::load(&filename).unwrap();
    let xs = Tensor::of_slice(&[1.0, 2.0, 3.0, 4.0, 5.0]);
    let ys = Tensor::of_slice(&[41.0, 1335.0, 0.1415, 4.0, 5.0]);
    let result = modl.method_ts("MyFn", &[xs, ys]).unwrap();
    assert_eq!(Vec::<f64>::from(&result), [42.0, 1337.0, 3.1415, 8.0, 10.0])
```

It's a bit unclear whether the current implementation (wrapping around the C++ `tracer::trace` function) is a good idea, re-implementing this entirely in Rust might be better.

### DIFF
--- a/src/wrappers/jit.rs
+++ b/src/wrappers/jit.rs
@@ -580,7 +580,7 @@ impl CModule {
         Ok(v)
     }
 
-    unsafe extern "C" fn traced_fn<F>(
+    extern "C" fn traced_fn<F>(
         user_data: *mut c_void,
         inputs: *const *mut C_tensor,
         ninputs: c_int,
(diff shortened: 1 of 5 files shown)


LaurentMazare:
I've pushed an alternative implementation that should avoid having to specify the number of outputs, and should be better at handling errors. The main drawback is that it uses more of the tracer internals so might end up being harder to maintain. The same test passes well so I'll merge this version.

PR #321, opened 2021-01-26
Add fallible (f_) counterparts to the Path initializers

danieldk:
In one of my projects, I am trying to go from unwrap-and-panic Torch errors to more user-friendly errors. This means that I have to bubble up all errors that come from `tch`, since errors can happen pretty much anywhere (e.g. when the GPU is OOM).

One major blocker is currently variable initialization using `Path`, since the initialization functions all call `unwrap()`ing functions. Failures often occur during variable initialization when there is some issue with CUDA.

This PR add fallible counterparts to the initializers in `Path` (prefixed with `f_`, like the fallible `Tensor` methods).

### DIFF
--- a/src/nn/init.rs
+++ b/src/nn/init.rs
@@ -1,5 +1,5 @@
 //! Variable initialization.
-use crate::{Device, Kind, Tensor};
+use crate::{Device, Kind, TchError, Tensor};
 
 /// Variable initializations.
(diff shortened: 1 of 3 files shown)


LaurentMazare (approved):

LaurentMazare:

PR #328, opened 2021-03-05
Refine excluded AMP prefix

danieldk:
Refine AMP function exclusion: `_amp` -> `_amp_foreach`

Exclusion of the `_amp` prefix also excludes `internal_amp_update_scale`,
which is needed for mixed-precision training.

### DIFF
--- a/gen/gen.ml
+++ b/gen/gen.ml
@@ -51,7 +51,7 @@ let prefixed_functions =
     (module String)
     [ "add"; "add_"; "div"; "div_"; "mul"; "mul_"; "sub"; "sub_"; "nll_loss"; "to_mkldnn" ]
 
-let excluded_prefixes = [ "_thnn_"; "_th_"; "thnn_"; "th_"; "_foreach"; "_amp" ]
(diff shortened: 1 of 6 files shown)


LaurentMazare (approved):

LaurentMazare:
Looks good, thanks for the PR! (though it would seem better if there is a function to do this that is not flagged as "internal" but it doesn't seem to be the case)

lostmsu:
Is there a example for using AMP?

joverwey:
I have the same question. How do you train with mixed precision. PyTorch has a class called `torch.cuda.amp.GradScalar`. What is the equivalent in `tch-rs`?

danieldk:
Here's an example of gradient scaling in Rust:

https://github.com/tensordot/syntaxdot/blob/main/syntaxdot/src/optimizers/grad_scale.rs

You also have to use `autocast` in forward passes:

https://github.com/tensordot/syntaxdot/blob/25a0f1893ef8107b2510b05cf977533db06affee/syntaxdot-cli/src/util.rs#L32

Issue #234, opened 2020-07-28
Cuda Stream Support


k21:
Thank you for creating and maintaining this library, it's great being able to experiment with machine learning in Rust. I am looking for potential performance improvements of some code that uses it.

Based on the **CUDA streams** section at https://pytorch.org/docs/stable/notes/cuda.html, it is my understanding that using streams is necessary to allow multiple operations to execute concurrently on a single GPU (one workaround could be to run operations from different processes, but that also has overhead).

Here are some sources I found that describe how streams can be used with the C++ API:
  * https://discuss.pytorch.org/t/cuda-streams-in-c-lib/38217
  * https://github.com/pytorch/pytorch/issues/16614

Would it be possible to add support for CUDA streams to tch?

NOBLES5E:
Related C++ struct: `at::cuda::CUDAStreamGuard`.

njaard:
Right now I'm loading your great `rust-bert` crate to run a model. I'd like to have a bunch of Rust threads do operations on the same model but it seems that I gradually run out of memory. I could wrap the model (and therefor all of torch) in a mutex, but then I can't saturate the GPU's throughput. I suspect that if I put every rust thread in its own Stream it would work. Do you agree with that?

If so, what are the steps for getting this feature merged? I could start by actually testing it!

freemanliu:
+1 for this feature. 
@LaurentMazare do you accept cxx bridge calling into libtorch?

pineapplemachine:
+1, I'm also wishing for CUDA stream support at the moment

PR #338, opened 2021-04-06
Preliminary support for stream guard.

LaurentMazare:
This adds some basic support for cuda streams #234.
Note that this has not been tested yet and is even unlikely to compile as I don't have access to a cuda enabled box at the moment.

Usage would be as follow.
```rust
let stream = tch::CudaStream::from_pool(/*high_priority=*/false, /*device_idx=*/1);
let _guard = tch::CudaStreamGuard::new(stream);
// The operations should here and until the end of the scope should take place in the created stream.
```


LaurentMazare:
There are still some difficulties to get this merged.
- The PyTorch C++ headers this require have references to the standard cuda headers.This [PyTorch issue](https://github.com/pytorch/pytorch/issues/55454) has been opened to see if we can get around this. In the meantime adding some `.include("/usr/lib/cuda/include")` in `torch-sys/build.rs` might help.
- This may require linking with c10_cuda, e.g. by adding the following line to `torch-sys/build.rs`.
```
println!("cargo:rustc-link-lib=c10_cuda");
```

LaurentMazare:
It may also be required to link `cudart` explicitely.

blogle:
Hey, curious if there is any movement on adding the stream api to tch-rs. I am working on an inference pipeline where having control over the cuda stream would be beneficial for eeking out some more performance.

LaurentMazare:
Not much progress I'm afraid. The change in this PR is pretty straightforward but the trickiness is that the required headers end up depending on the cuda headers and these headers may not be installed if cuda was vendored with pytorch rather than properly installed. The related PyTorch issue https://github.com/pytorch/pytorch/issues/55454 hasn't seen much activity, (also related is https://github.com/pytorch/pytorch/issues/47743 ).
Having these headers available would unlock both support of cuda stream and cuda graph for `tch` so would be great to have but it's a bit stuck at the moment.

LaurentMazare:
Vendoring the cuda headers might be a bit tricky. A specific version of `tch` is tied to a specific release of `libtorch` but these may used different versions of cuda (which I guess is pretty helpful to users with older gpus/nvidia drivers).
No clue how much effort it would be to host fully self-contained binaries, but I guess it might be non-trivial as these would have to run in a wide variety of environments - maybe there are some specific bits of the PyTorch ecosystem that could be leveraged for this e.g. [pytorch/builder](https://github.com/pytorch/builder).
Another possibility would be for `tch` to try and detect the cuda install and only activate the advanced cuda functionalities (cuda guards and all) if these are available.

Overall the best would certainly be if the `libtorch` binaries could embed a bit more of the header files but indeed there hasn't been much progress on this over the last year so we should probably find a good way around this.

PR #339, opened 2021-04-08
Use windows unicode paths to get around windows 260 char path limit.

MarkJGx:
`libtorch` fails to extract on Windows when the project is near the 260 char windows path limit, this adds a windows check and adjusts the path. Secondly, this normalizes paths using `to_verbatim`, normalizing paths should fix any issues while using a custom `CARGO_TARGET_DIR` on Windows. This PR still needs some work.

### DIFF
--- a/torch-sys/Cargo.toml
+++ b/torch-sys/Cargo.toml
@@ -20,6 +20,9 @@ anyhow = "1.0"
 cc = "1.0"
 curl = "0.4.9"
 zip = "0.5"
+normpath="0.3.0"
+verbatim="0.1.1"
+cfg-if="1.0.0"
 
 [features]
 doc-only = []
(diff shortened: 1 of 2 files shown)


MarkJGx:
Accidentally closed this. Anyhow while this does fix the extraction issues on windows by using the long unicode paths, it reveals a new issue. MSVC's compiler driver cl.exe fails to find certain include headers when building, which coincidentally are longer than 260 characters on my windows file system. I'm not sure if there's a flag that could be passed through `cc` that could fix this, might be an unresolved `cl` issue. 

I'm currently working with a much older commit (0.3.1), I've made changes that provide linux function fallbacks without looking as bad as this PR with the cfg macros. Most likely will forward port later when I've resolved the cl issues.

PR #340, opened 2021-04-08
Mark Element trait depend on Copy and mark it as unsafe.

ggoossen:
Since Element is used to copy elements from a tensor, it should have the copy trait, and is unsafe because the size must match.

Also add some documentation. Mentioning Element from Kind.

### DIFF
--- a/src/wrappers/kind.rs
+++ b/src/wrappers/kind.rs
@@ -3,6 +3,8 @@
 use half::f16;
 
 /// The different kind of elements that a Tensor can hold.
+///
+/// [Element] can be used for a value of a specific kind.
 #[allow(clippy::upper_case_acronyms)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub enum Kind {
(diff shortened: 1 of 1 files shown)


edlanglois:
@LaurentMazare bumping this with an example of using the `Element` trait to read unitialized/arbitrary memory without `unsafe`
```rust
use tch::{kind::Element, Device, Kind, Tensor};

#[derive(Clone, Debug)]
struct X(u64);

impl Element for X {
    const KIND: Kind = Kind::Int8;
    const ZERO: Self = Self(0);
}

let data: Vec<X> = Tensor::ones(&[10], (Kind::Int8, Device::Cpu)).into();
println!("{:?}", data);
```
gives `[X(72340172838076673), X(257), X(0), X(0), X(0), X(0), X(0), X(0), X(0), X(0)]`

@ggoossen it's not just the size that has to match but also the data type. I feel like any comment should just say not to implement for other types.

LaurentMazare:
Yeah, I agree that it seems right to mark this at unsafe, thanks for pointing it out. @ggoossen if you're still interested would you mind rebasing/fixing the merge issues? If you don't have the time, I'll go over it in the next couple days.

LaurentMazare:
Just pushed some changes to mark the trait as unsafe, sorry for not looking at this earlier.

PR #347, opened 2021-04-28
Fix build on Jetson Xavier AGX

elaye:
Hello,
I was trying to build `tch` on the Jetson Xavier AGX board (arm64) and I got two errors.
This PR fixes the issue in case you're interested.

### DIFF
--- a/src/wrappers/jit.rs
+++ b/src/wrappers/jit.rs
@@ -346,7 +346,7 @@ impl IValue {
             }
             8 => {
                 let len = unsafe_torch_err!(ati_length(c_ivalue));
-                let mut c_array = vec![0_i8; len as usize];
+                let mut c_array = vec![0_u8; len as usize];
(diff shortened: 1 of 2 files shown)


LaurentMazare:
Thanks for the PR, the bool_to_option feature in `build.rs` seems to break the CI.
Are you running with the stable channel?

elaye:
Ah, yes sorry I was building on nightly.
I removed the feature flag so the CI should pass on stable now.

elaye:
So it seems that `char` is interpreted differently on ARM. I added a cast to `libc::c_char` that should do the trick.

However I can't compile on my machine without adding `#![feature(bool_to_option)]` to `torch_sys::build.rs` and I don't understand why, if it passes the CI without it.

LaurentMazare:
The bool_to_option feature indeed seems mostly unrelated so not sure what is going on, the CI also runs on nightly and doesn't seem to hit the issue. Anyway I'll merge the PR, thanks!

LaurentMazare (approved):

PR #351, opened 2021-05-12
Updated gradient_descent example

nonnominandus:
We ran into some problems in running the `gradient_descent` example, these changes made it possible to run the example for us. 

(Worked with @chrisalgerges98)

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,8 +96,8 @@ Then on each step of the training loop:
 
 ```rust
 extern crate tch;
-use tch::nn;
-use tch::Tensor;
+use tch::nn::*;
+use tch::*;
 
 fn my_module(p: nn::Path, dim: i64) -> impl nn::Module {
     let x1 = p.zeros("x1", &[dim]);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,8 +96,8 @@ Then on each step of the training loop:
 
 ```rust
 extern crate tch;
-use tch::nn;
-use tch::Tensor;
+use tch::nn::*;
+use tch::*;
```
LaurentMazare, line 100: Maybe we could avoid the wildcard uses as they bring lots of things in the scope and instead a `use tch::kind` would be enough?

nonnominandus:
Leaving out all wildcards would give the following imports. 
```rust
use tch::kind;
use tch::nn;
use tch::nn::Module; // for the `forward` method
use tch::nn::OptimizerConfig; // for the `build` method
use tch::Device;
use tch::Tensor;
```
If you feel this is a better way to do it I will change my PR.

LaurentMazare:
I would slightly lean towards something like this indeed. Note that you can make things a bit more compact if you prefer, e.g.:
```rust
use tch::{kind,nn,Device,Tensor};
use tch::nn::{Module,OptimizerConfig};
```

nonnominandus:
Great, it is pushed to this PR. Thanks for the reminder of the compact form of the imports.

LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR!

PR #356, opened 2021-05-19
Use generics instead of macros for From, TryFrom, TryInto impls

edlanglois:
In most cases the macros were not necessary and can be replaced by a generic implementation over `T: crate::kind::Element`.

The `T: Default` bound is to allow zero-initializing Vec in `From<&Tensor> for Vec<T>`.
There's probably an unsafe way to avoid the zero-initialization since those values should be overwritten anyway, but that's out of scope for this commit.

### DIFF
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -483,35 +483,73 @@ impl Tensor {
     }
 }
 
-macro_rules! from_tensor {
-    ($typ:ident, $zero:expr, $kind:ident) => {
-        impl From<&Tensor> for Vec<$typ> {
-            fn from(tensor: &Tensor) -> Vec<$typ> {
-                let numel = tensor.numel();
-                let mut vec = vec![$zero; numel as usize];
-                tensor.to_kind(Kind::$kind).copy_data(&mut vec, numel);
-                vec
-            }
-        }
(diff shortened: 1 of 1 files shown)


edlanglois:
BTW the main thing I wanted with these changes is `TryFrom<&Array>`. The logic for the rest of them is that
* If implementing `TryFrom<&Array>` then one might as well do `TryFrom<&ArrayBase>` instead since it is more generic
* `TryFrom<&`ArrayBase>` needs to use generics, but adding the appropriate generics to the existing macro definitions doesn't work because they conflict with each other (at least, as far as the compiler is concerned) so the whole thing needs to be a macro.

These changes should be backward compatible with the only effects being adding `TryFrom<&Vec>` and `TryFrom<&Array>` / `TryFrom<&ArrayBase>`, possibly with support for some more types if they implement `kind::Element` but weren't listed in the macro calls. I also agree with the changes in #297, which are still relevant if this is applied.

LaurentMazare:
I just pushed some long overdue splitting of the tensor module in #372 and in the process I switched some of the conversion macros to be using generics intead + used `ArrayBase` instead of `Array` so hopefully this should cover your use cases. Apologies for the delay.

edlanglois:
Nice! No worries at all about the delay.

PR #357, opened 2021-05-19
Add TryFrom<&Vec> and TryFrom<&Array>

edlanglois:
The existing `TryFrom<Vec>` and `TryFrom<Array>` only need references so this just changes the original signatures to use references and adds new wrapper implementations for `TryFrom<Vec>` and `TryFrom<Array>`.

This builds on top of #356 (although it can be implemented without that change) so all of the changes from #356 are included in this diff as well. I don't know how to prevent that but anyway, don't merge this unless #356 has been merged.

### DIFF
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -724,14 +724,14 @@ where
     }
 }
 
-impl<T, D> TryFrom<ndarray::Array<T, D>> for Tensor
+impl<T, D> TryFrom<&ndarray::Array<T, D>> for Tensor
 where
     T: crate::kind::Element,
     D: ndarray::Dimension,
 {
     type Error = TchError;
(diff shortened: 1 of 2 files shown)


jerry73204:
@edlanglois We should implement `TryFrom` instead of `TryInto`. Says by [std doc](https://doc.rust-lang.org/std/convert/trait.TryInto.html), `TryFrom` implies `TryInto` but it does not in vice versa. A similar fix can be found in my old #297 PR.

edlanglois:
I agree with that. For these commits I tried to keep each one doing as small a change as possible so I figure your existing PR covers the case of changing to `TryFrom` or a new PR could be made with just `TryInto => TryFrom` if these are accepted.

PR #359, opened 2021-05-19
Use ok_or_else in `TryFrom<&ArrayBase>`

edlanglois:
* Use `ok_or_else` instead of match
~~* Return value directly instead of separate `let v = ...?; Ok(v)`~~

~~This builds on #356 (but can be implemented without it). Don't merge unless #356 has been merged.~~


edlanglois:
This is the only relevant change left from this PR. `ok_or_else` seemed like an appropriate alternative to waiting for `NoneError` but it's pretty minor so feel free to close if you prefer the existing code.

LaurentMazare:
Merged, seems slightly more idiomatic indeed.

PR #361, opened 2021-05-20
Fix lifetime annotation of some Path operations.

cyb0124:
Consider this line: `let path2 = path1.sub("test")`

Currently, because of a lifetime annotation, compiler consider `path2` be borrowing `path1`. This prevents `path1` being dropped while `path2` is in use. However, this is unnecessary as `path2` isn't actually borrowing anything in `path1`. This change removes the annotation.


LaurentMazare (approved):

LaurentMazare:

Issue #360, opened 2021-05-20
Saving optimizer state


cyb0124:
It doesn't seem like optimizers' internal states are stored in `VarStore`. I also don't find any public interfaces for explicitly accessing / saving optimizers' internal states. So, is it possible to save and reload training state without losing optimizers' internal state?

KouAkitaki:
I've taken a look into this. From the C++ API, I see [this `torch::save` function template](https://pytorch.org/cppdocs/api/function_namespacetorch_1a99dc9f736064b2179cc58e6436f7a021.html?highlight=save) that accepts `torch::optim::Optimizer` as input, but it's not present in `torch-sys` API, probably since the generator does not generate it. We'll first need to add it to `torch-sys/libtch/torch_api.{cpp,h}`, and expose it in `torch-sys/src/lib.rs`, in order for `tch` to use it.

PR #388, opened 2021-07-15
Add support for saving and loading optimizers

KouAkitaki:
Addresses #360. Not entirely sure how to test it, since the `Optimizer` api is quite opaque, making it hard to determine if `load` actually loads up states correctly.

### DIFF
--- a/src/nn/optimizer.rs
+++ b/src/nn/optimizer.rs
@@ -228,6 +228,16 @@ impl<T> Optimizer<T> {
         self.opt.step().unwrap()
     }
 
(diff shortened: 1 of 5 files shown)


lostmsu:
The way I usually test my checkpointing is plot the loss, checkpoint and resume, then ensure loss curve does not have spikes at the resume point.

pragmaticTNT:
@LaurentMazare I was looking to implement a similar functionality. If I submitted a pr, would it have a chance to get merged and what differences should it have to this one?

PR #394, opened 2021-07-29
vision: Add APIs to load images from memory

arilou:
Signed-off-by: Jon Doron <arilou@gmail.com>

### DIFF
--- a/src/vision/image.rs
+++ b/src/vision/image.rs
@@ -1,5 +1,5 @@
 //! Utility functions to manipulate images.
-use crate::wrappers::image::{load_hwc, resize_hwc, save_hwc};
+use crate::wrappers::image::{load_hwc, load_hwc_from_mem, resize_hwc, save_hwc};
 use crate::{Device, TchError, Tensor};
 use std::io;
 use std::path::Path;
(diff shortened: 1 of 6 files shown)


LaurentMazare:
Could you give a bit more details about what this PR achieves? In particular I'm wondering why one would use this rather than creating a tensor directly from the memory and reshaping it? (though maybe some helper functions could be useful for this)

arilou:
Sure, basically I'm getting a stream of PNG files which I wanted to classify, so because I already had them as a buffer I preferred to just load them from there rather than writing to a file and then doing load_image_and_resize224.

You are saying there is a different alternative to this? (I simply looked at the libtorch API and saw you simply did not wrap it)

LaurentMazare (approved):

LaurentMazare:
The [stbi_...] functions are actually not part of libtorch but instead are from another library, some of the already exposed libtorch functions are likely to already do the same kind of thing (the wrappers for these are automatically generated). Anyway it seems reasonable to also expose this as it provides a more uniform user api.
Would you mind running rustfmt on your changeset? This breaks the CI testing at the moment.

arilou:
Thanks I have run cargo fmt

LaurentMazare:
Merged, thanks.

PR #411, opened 2021-09-05
Add Tensor to ArrayView(Mut) conversion and minor improvements

makroiss:
- Use uninitialized vectors
- Add Tensor to ArrayView(Mut) conversion
- Extend Tensor to Array conversion for non dynamic dimensions

### DIFF
--- a/src/tensor/convert.rs
+++ b/src/tensor/convert.rs
@@ -1,13 +1,16 @@
 //! Implement conversion traits for tensors
 use super::Tensor;
-use crate::{kind::Element, TchError};
+use crate::{kind::Element, Device, TchError};
 use half::f16;
-use std::convert::{TryFrom, TryInto};
+use std::convert::TryFrom;
 
 impl<T: Element> From<&Tensor> for Vec<T> {
     fn from(tensor: &Tensor) -> Vec<T> {
         let numel = tensor.numel();
-        let mut vec = vec![T::ZERO; numel as usize];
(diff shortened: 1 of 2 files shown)


edlanglois:
Is this safe? What happens if you create a view on a tensor `t` then call `t.resize_`?

makroiss:
Yes, good point. If the buffer changes without creating a new tensor handle, then this is unsafe. I found this useful to avoid copies but happy to revert if this is too dangerous.

edlanglois:
I wanted this functionality too so you inspired me to try creating a hopefully-safe version: https://gist.github.com/edlanglois/5444813b409a9b44f4067591bf0f40b0

All views are provided through `ExclusiveTensor`, which creates the tensor and keeps it private so that it has exclusive access to the tensor data. `ExclusiveTensor` can then be converted into a regular `Tensor`.

This is my first attempt at writing unsafe code and I'm not super familiar with how Tensor works on the C++ side, so it is still possible that I missed something and this is unsafe.

makroiss:
Nice, a problem might be that the ArrayView can outlive the ExclusiveTensor. You should be able to solve this with lifetime annotations similar to the code I sent.

edlanglois:
I think that part is OK, there is an elided lifetime in the returned `ArrayView` so the signature is actually `array_view<'a>(&'a self) -> ArrayView<'a, E, D>` and it can't outlive `ExclusiveTensor`. Attempting to use the view after `into_tensor()` is a compile error.

Issue #415, opened 2021-09-09
Module-like methods for VarStore


guillaume-be:
Hello,

Based on https://github.com/guillaume-be/rust-bert/issues/183, I am currently thinking of an effective way of implementing module-level conversion to and from half-precision. Pytorch's model are based on an object-oriented approach where a `nn.Module` stores both the model weights and associated forward method. This allows the user to call conversion or device placement method directly at the module level:
```python
class NeuralNet(nn.Module)
    def __init__(self):
        self.layer = nn.Linear(10, 42)

    def forward(self, x):
         return self.layer(x)

model = NeuralNet()
model = model.half()
model = model.cuda(0)
```

This is particularly useful when working with large models, where at prediction-time the process would be to:
1. Instantiate the model on the CPU/memory and load weights in full precision
2. Cast the model to half-precision
3. Move the model to the GPU buffer

In _tch_ the model behaviour (for example implementing `ModuleT`) is separated from the parameter storage (in a `VarStore`). I could not find methods to perform the half-precision conversion at an entire module level or to move a module between devices. I believe this could be done by extending the current `VarStore` or `Path` with additional methods:
1. `to_device(device: Device)`: move a `VarStore` or a `Path` to the given device. possibly extended with `cpu()` and `cuda(usize)` helpers
2. `half()`: converts all registered `Tensor` in a `Path` to half-precision if the tensor is of floating-point type, see https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module)
2. `bloaft16()`: converts all registered `Tensor` in a `Path` to bfloat16 half-precision if the tensor is of floating-point type
3. `float()`: converts all registered `Tensor` in a `Path` to single precision if the tensor is of floating-point type
4. `double()`: converts all registered `Tensor` in a `Path` to double precision if the tensor is of floating-point type

Do you think this would be a reasonable approach to implement the Pytorch's `nn.Module` utilities? If so I'd be happy to work on an update for your review to enable convenient casting to half precision and model device placement.

LaurentMazare:
Thanks for the detailed write-up.
Indeed a var-store is attached to a device, but it's not really attached to a *Kind* (float/double/half) at the moment as mostly everything is done with float. Having only one Kind for a var-store is an interesting idea and probably an improvement over the current setup but I'm wondering how common it is for models to use mixed precision in a way this would be a problem here.
[...]

PR #416, opened 2021-09-11
VarStore extensions

guillaume-be:
Hello,

This PR is a first attempt of implementation of the ideas discussed in https://github.com/LaurentMazare/tch-rs/issues/415.
I implemented:
- VarStore-level casting between the different precision levels for Float types
- VarStore-level migration between devices
- Path-level casting between the different precision levels for Float types. 

Please let me know what you think

### DIFF
--- a/src/nn/var_store.rs
+++ b/src/nn/var_store.rs
@@ -196,6 +196,54 @@ impl VarStore {
         }
     }
 
+    /// Casts all float-like variable of a var store to half-precision (Half kind).
+    pub fn half(&mut self) {
+        let mut variables = self.variables_.lock().unwrap();
+        for (_, variable) in variables.named_variables.iter_mut() {
+            if variable.is_floating_point() {
+                *variable = variable.to_kind(Kind::Half);
+            }
+        }
+    }
+
+    /// Casts all float-like variable of a var store to bfloat16-precision (BFloat16 kind).
+    pub fn bfloat16(&mut self) {
+        let mut variables = self.variables_.lock().unwrap();
+        for (_, variable) in variables.named_variables.iter_mut() {
+            if variable.is_floating_point() {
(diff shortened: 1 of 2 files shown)


src/nn/var_store.rs
```diff
@@ -265,6 +294,62 @@ impl<'a> Path<'a> {
...
+    /// other VarStore variables are unaffected
+    pub fn half(&mut self) {
+        let path_root = self.path.join(SEP.to_string().as_str());
+        let mut variables = self.var_store.variables_.lock().unwrap();
+        for (variable_name, variable) in variables.named_variables.iter_mut() {
+            if variable_name.starts_with(&path_root) & variable.is_floating_point() {
+                *variable = variable.to_kind(Kind::Half);
+            }
+        }
+    }
```
LaurentMazare, line 309: Would it be possible to have a function with a kind argument: it would avoid duplicating the code 4 times and also could be exposed to the end users if they want to use a kind argument.

guillaume-be: Thank you for the suggestion. I have added a new `to_kind` method exposed to the user that would allow to convert a path/var store to any `Kind`. I would rely on a different (private) `to_float_kind` for the implementation of `half`, `float`, `bfloat16` and `double` to only process float-like variables in the var store. This would be matching Pytorch's implementation at https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.half

LaurentMazare (commented):
Thanks, looks pretty good, there is a clippy lint though that fails in the CI test mentioning to_* functions using mut self being an issue. Maybe naming them set_* would both fix the lint and make it more explicit that this works in place.

LaurentMazare (approved):

LaurentMazare:
Thanks for the PR!

Issue #421, opened 2021-09-27
Handling of optional bias


guillaume-be:
Hello,

Following https://github.com/LaurentMazare/tch-rs/pull/420 the type and device conversion of var stores (or sub-paths of it) seem to be working properly, and I managed to run a few models in fp16 mode with this new feature. 

I encountered one issue with the linear layers when the `bias` was set to `false`. The current implementation then creates an internal `Tensor` of zeroes (which is not registered to the provided path) and adds it to the output of the weights-input matrix multiplication. This breaks when converting between type sand devices because the "dummy" bias tensor does not get converted (not registered in the var store).

I updated my local copy of `tch-rs` to make the bias truly optional (and avoid a dummy operation when set to false), along the lines of:
```rust
/// A linear fully-connected layer.
#[derive(Debug)]
pub struct Linear {
    pub ws: Tensor,
    pub bs: Option<Tensor>,
}

/// Creates a new linear layer.
pub fn linear<'a, T: Borrow<super::Path<'a>>>(
    vs: T,
    in_dim: i64,
    out_dim: i64,
    c: LinearConfig,
) -> Linear {
    let vs = vs.borrow();
    let bs = if c.bias {
        let bs_init = c.bs_init.unwrap_or_else(|| {
            let bound = 1.0 / (in_dim as f64).sqrt();
            super::Init::Uniform {
                lo: -bound,
                up: bound,
            }
        });
        Some(vs.var("bias", &[out_dim], bs_init))
    } else {
        None
    };

    Linear {
        ws: vs.var("weight", &[out_dim, in_dim], c.ws_init),
        bs,
    }
}

impl super::module::Module for Linear {
    fn forward(&self, xs: &Tensor) -> Tensor {
        if let Some(bias) = &self.bs {
            xs.matmul(&self.ws.tr()) + bias
        } else {
            xs.matmul(&self.ws.tr())
        }
    }
}
```
I believe this is closer to the [C++ implementation](https://github.com/pytorch/pytorch/blob/facff2ec6562d61eab2ccd9f1c16872e54eec92d/aten/src/ATen/native/Linear.cpp#L37) and to the `Conv` block defined in this crate. 

I'd be happy to contribute these changes back - but I wanted to check with you beforehand as this would cause breaking changes for dependents manipulating the bias field in `Linear` directly. Unfortunately, registering the `bias` field to the var store would solve the issue without causing breaking change, but would raise errors when loading weight files where this "dummy" bias is not saved.

[...]

PR #422, opened 2021-09-28
Updated `Linear` bias to be optional instead of zeroes

guillaume-be:
- Change type of `bs` to `Option<Tensor>`
- Updated forward pass to add bias only when non-None
- Updated unit tests that were accessing the `bs` field directly

Fixes https://github.com/LaurentMazare/tch-rs/issues/421


LaurentMazare (approved):

LaurentMazare:
Looks good thanks, fwiw it seems that convolutions were already working this way so this change should make things more uniform.

Issue #438, opened 2021-12-24
Add support Read+Seek trait


tsubakisakura:
Hello!

I have tried restoring VarStore from ```std::io::File```, ```std::io::Cursor```, etc, but there was no direct loading API.
So, I thought it would be convenient if the following APIs were available in the VarStore.

~~~
pub trait ReadStream : Read+Seek {}

impl VarStore {
    ...
    pub fn load_from_stream<S: ReadStream>(&mut self, stream: S) -> Result<(), TchError>
    ...
}
~~~

libtorch has a ```torch::jit::load``` function that takes ```caffe2::serialize::ReadAdapterInterface```.
If I implemented this, it seemed to connect to the Rust side. I tried to implement it and it worked fine.
How about an API like this? If this API policy is acceptable, I will make a pull request.

Thanks.

LaurentMazare:
Hello,
Sounds like an interesting idea. Could you say a bit more about the potential use case for this, how the api would look like and what would have to be implemented? I just gave a quick look at `caffe2::serialize::ReadAdapterInterface` and it seems very generic so it's a bit unclear to me what aspects you would want to cover here.

tsubakisakura:
Thank you for your reply!

I have already implemented the required implementation code, so I pushed it.
https://github.com/tsubakisakura/tch-rs/commit/f4714da233ab2101174539a52a66b1555a7fa50b

My real purpose is to get the thread to do the job as much as possible.
That's why I wanted to run disk I/O in a separate thread.
The thread does not need disk access if ```std::io::Cursor``` is available.

But I realized there might be another way.
The I/O thread can use ```Tensor::load_multi``` if the VarStore can be initialized with ```Vec<(String,Tensor)>```.
This seems to be more efficient because the deserialization process can be done by the I/O thread.

LaurentMazare:
Thanks, the change looks fairly reasonable, happy to get some PR with it. Two small points:
- Could the `at_load_from_stream_callback_with_device` and `at_load_from_stream_callback` C++ functions be shared, e.g. using a device of -1 to represent the case where no device was provided?
- Could the `extern "C"` bits be moved to `torch_api.h` rather than `torch_api.cpp` (similar to all the other functions)?

tsubakisakura:
Thank you for your feedback and reviews!

Since ```Device::Cpu.c_int()``` is also -1, ```at_load_from_stream_callback``` cannot determine whether the default argument or ```Device::Cpu``` was specified.
[...]

PR #441, opened 2022-01-04
Add Read+Seek trait and Write trait support

tsubakisakura:
see also: https://github.com/LaurentMazare/tch-rs/issues/438

### DIFF
--- a/torch-sys/src/io.rs
+++ b/torch-sys/src/io.rs
@@ -21,7 +21,7 @@ extern "C" fn tch_write_stream_write(
     unsafe {
         let boxed_stream = stream_ptr as *mut Box<dyn Write>;
         let buffer = std::slice::from_raw_parts(buf, size);
-        match (*boxed_stream).write(&buffer) {
+        match (*boxed_stream).write(buffer) {
             Ok(x) => {
                 *out_size = x;
                 true
--- a/torch-sys/src/io.rs
+++ b/torch-sys/src/io.rs
@@ -3,11 +3,6 @@ use std::io::{Read, Seek, SeekFrom, Write};
 
 pub trait ReadStream: Read + Seek {}
 
-impl ReadStream for std::fs::File {}
-impl ReadStream for std::io::Cursor<Vec<u8>> {}
-impl ReadStream for std::io::Cursor<&[u8]> {}
-impl<T: Read + Seek> ReadStream for std::io::BufReader<T> {}
-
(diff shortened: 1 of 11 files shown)


torch-sys/src/io.rs
```diff
@@ -0,0 +1,80 @@
+use libc::{c_void, size_t};
+use std::io::{Write, Read, Seek, SeekFrom};
+
+pub trait ReadStream : Read+Seek {}
```
LaurentMazare, line 4: Is there a reason for adding a new trait rather than relying on the existing `Read` and `Seek` one? Having a different trait here makes it harder for user to use this on types that are not defined in their crate and that would implement both `Read` and `Seek` (they can wrap the type but this doesn't sound ideal).

torch-sys/src/io.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl ReadStream for std::fs::File {}
+impl ReadStream for std::io::Cursor<Vec<u8>> {}
+impl ReadStream for std::io::Cursor<&[u8]> {}
+impl<T:Read+Seek> ReadStream for std::io::BufReader<T> {}
+
+#[no_mangle]
+extern fn tch_write_stream_destructor(stream_ptr: *mut c_void) {
+    unsafe {
+        let _ : Box<Box<dyn Write>> = Box::from_raw(stream_ptr as *mut Box<dyn Write>);
```
LaurentMazare, line 14: Calling `drop(Box::from_raw(...))` might be slightly clearer than `let _ = `.

LaurentMazare:
Could you please run `rustfmt` and `clippy`? This would help getting the CI happy.

tsubakisakura:
I have created an internal adapter that implements ReadStream so that Read + Seek is sufficient as an external interface. how is it?
(The ReadStream trait is needed to erase the type information, so I decided to use it only internally.)

LaurentMazare:
Sorry for the delay, looks pretty good to me except that rustfmt seems to fail in the CI. Would you mind running it again?

tsubakisakura:
I'm sorry, I didn't notice that the latest branch ```rustfmt.toml``` has been updated.
Updated and rebased.

LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR.

tsubakisakura:
Thank you for merging! 😀

PR #444, opened 2022-01-17
lstm/gru: use Borrow<Path> instead of &var_store::Path

slckl:
This changes `nn::rnn::lstm()` and `nn::rnn::gru()` function signatures to take in a `Borrow<Path>` instead of a `&Path`, similar to how it's done seemingly everywhere else.

Without this, lstm/gru creation is slightly less ergonomic than other layers. Example of inconsistency:

```Rust
let embed = nn::embedding(
    // we can easily use "/" for custom path and don't have to reborrow whole thing manually
    &vs.root() / "embed",
    ..
);
let lstm = nn::lstm(
    // have to manually borrow the result of the "/" expression...
    &(&vs.root() / "lstm"),
    ..
);
```

This should be backwards-compatible.

### DIFF
--- a/src/nn/rnn.rs
+++ b/src/nn/rnn.rs
@@ -1,5 +1,6 @@
 //! Recurrent Neural Networks
 use crate::{Device, Kind, Tensor};
+use std::borrow::Borrow;
 
 /// Trait for Recurrent Neural Networks.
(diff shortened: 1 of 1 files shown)


LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR!

Issue #434, opened 2021-12-02
example/mnist/mnist_conv.rs run failed 


ChenHuaYou:
if i run the example in main branch i would failed.    
 Running `target/debug/examples/mnist conv`
[...]

PR #446, opened 2022-01-20
Patch mnist_conv example (see #434)

daviddao:
Currently in the mnist examples, `mnist_conv.rs` panics because `dropout_` is changing a value in-place. Changing it to `dropout` fixes the issue (see #434)


LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the fix!

PR #448, opened 2022-01-22
Use Iter2's device instead of hardcoding CPU for shuffle method

ViliamVadocz:
Previously this would crash if `self.xs` or `self.xy` was on CUDA because it the tensor used in `index_select` was on a different device.

All tests pass.
I have tested this with and without CUDA and it doesn't crash.
I have not checked the correctness of the output, but I assume it is correct, since it is such a small change I don't see how it could go wrong.


LaurentMazare (approved):

LaurentMazare:
Merged, thanks for catching and fixing this!

PR #450, opened 2022-01-24
use std::fs::read instead of manual BufReader in data::TextData::new()

slckl:
hi, sry, another drive-by "improvement"

`TextData::new()` function manually creates a `BufReader` around `File`, and then calls `read_to_end` on it to read all the data in memory. In `read_to_end` case, `BufReader` actually defers to inner's implementation, in this case `File`'s impl, we can easily verify this here: https://doc.rust-lang.org/stable/src/std/io/buffered/bufreader.rs.html#313

But for handling `File::read_to_end()` into a `Vec<u8>`, Rust's std library has a nice convenience method ready to go: `std::fs::read()`

This PR merely calls the stdlib convenience function for reading file into vec of bytes. This should have no measurable performance impact. The code is a bit shorter, though.

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -1,8 +1,6 @@
 //! Dataset iterators.
 use crate::{kind, kind::Kind, Device, IndexOp, TchError, Tensor};
 use std::collections::HashMap;
-use std::fs::File;
-use std::io::{BufReader, Read};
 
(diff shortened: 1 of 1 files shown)


LaurentMazare (approved):

LaurentMazare:
Looks good, thanks for the PR.

PR #451, opened 2022-01-25
Support TorchScript objects

AlexanderEkdahl:
This PR allows interacting with TorchScript Classes:

```
@torch.jit.script
class TorchScriptClass:
    def __init__(self, x: torch.Tensor):
        self.x = x

    def y(self) -> torch.Tensor:
        return self.x * 2

@torch.jit.script
def foo_6(x: torch.Tensor):
    return TorchScriptClass(x)
 ```
 
Currently, the user would encounter `unsupported tagObject` when running any model that returns a TorchScript class.

### CODE REVIEW

File: src/wrappers/jit.rs [hunk 1]
```diff
@@ -9,6 +9,7 @@ use torch_sys::*;
 
 /// Argument and output values for JIT models.
 #[derive(Debug, PartialEq)]
+#[non_exhaustive]
```
AlexanderEkdahl, line 12: https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute

torch-sys/libtch/torch_api.cpp
```diff
@@ -1280,6 +1281,17 @@ void ati_to_tensor_list(ivalue i,
   )
 }
 
+ivalue ati_object_method_(ivalue i, char *method_name, ivalue *ivalues, int nivalues) {
```
LaurentMazare, line 1284: Could we avoid using the same `i` name for the input ivalue and the for loop iteration below?

LaurentMazare:
Looks pretty good, I've just put one cosmetic comment in the review.

torch-sys/libtch/torch_api.cpp
AlexanderEkdahl: I creatively renamed it to `j`. I agree on the change but I was mostly following existing patterns which used `i`.

AlexanderEkdahl:
> Looks pretty good, I've just put one cosmetic comment in the review. Also `rustfmt` fails on the CI, could you ensure that it's run?

Sorry about that. I changed the variable name and ran `cargo fmt`.

LaurentMazare (approved):

LaurentMazare:
Thanks for making the changes, actually it seems that the new tests are not passing on macOS CI, would you mind taking a look at whether there is anything that could be done to fix this? [CI details](https://github.com/LaurentMazare/tch-rs/runs/5042183928?check_suite_focus=true)

AlexanderEkdahl:
```
---- save_and_load_from_stream_var_store stdout ----
thread 'save_and_load_from_stream_var_store' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', tests/var_store.rs:112:57
```

The failure doesn't look related to this change? I developed this feature on a Linux platform but I could take a look to see if I can reproduce on my Mac.

LaurentMazare:
Oh right, I got confused as this other PR landed a couple days ago, agreed that it doesn't seem related to this PR so I'll merge it.

PR #467, opened 2022-03-29
Expose set graph executor optimize.

vcarpani:
This PR exposes `torch::jit::setGraphExecutorOptimize()`.

I tested it and the behavior of the network changes (reduces the amount of warmups needed), but it is not as effective as in the original C++ implementation, so I am open to any suggestions on how to improve this

### DIFF
--- a/src/wrappers/jit.rs
+++ b/src/wrappers/jit.rs
@@ -701,6 +701,15 @@ pub fn set_profiling_mode(b: bool) {
     f_set_profiling_mode(b).unwrap()
 }
 
+pub fn f_set_graph_executor_optimize(b: bool) -> Result<(), TchError> {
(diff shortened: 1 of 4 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks for the PR, looks like there is a clippy lint failure in the CI, would mind fixing it?

LaurentMazare:
Merged, thanks for the PR!

de-vri-es:
Would it be possible to get a minor/patch (0.7.1) release for this feature? It would be really useful for us :)

LaurentMazare:
Sure, I've messed things up with 0.7.1 but 0.7.2 should now be available via crates.io and hopefully fine.

de-vri-es:
Awesome, thank you!

PR #468, opened 2022-03-30
add more dbtype for read_npy which already support by tensor

any35:
When working with numpy, I found that the full supported types of tensor are:
  float64, float32, float16,
  complex64, complex128,
  int64, int32, int16, int8,
  uint8, and bool.

So I add some code to support read all support dbtype of numpy files to tensor.


LaurentMazare:
Merged, thanks for the PR.

PR #470, opened 2022-04-01
Added an example that runs a custom trained classifier on webcam feed.

metobom:
I added a simple example that shows how to run our custom-trained PyTorch models on a webcam with OpenCV and tch-rs. The example runs on GPU.
* EDIT1: Current version works on a video. But since opencv-rust has the same usage as Python and C++, it should be easy to make it work on a webcam. 
* EDIT2: I added the command for running the example on webcam.


LaurentMazare:
Thanks for working on this, looks like a great demo. That said I'm not thrilled by the idea of adding a dependency to opencv (at least it should be a dev-dependency) nor by having the large binary example/model weights in the repo. What about instead you put this example in its own repo / create a separate crate for it? And we can certainly add a link to it from the main readme so that it's easy for users to discover.

metobom:
Here I created a repo: https://github.com/metobom/tchrs-opencv-webcam-inference

LaurentMazare:
Great! I've added a pointer to it with a short description in the main readme.

Issue #475, opened 2022-04-21
Segmentation fault in CModule.method_is with Object parameter


neilisaac:
I believe there may be a double-free issue in method_is operating on a borrowed Object: https://github.com/LaurentMazare/tch-rs/blob/main/src/wrappers/jit.rs#L518

Example use:
```python
@torch.jit.script
class InputObject:
    def __init__(self, foo, bar):
        self.foo = foo
        self.bar = bar

class TorchscriptExample(torch.nn.Module):
    @torch.jit.export
    def add_them(self, data: InputObject) -> torch.Tensor:
        return data.foo + data.bar

    @torch.jit.export
    def make_input_object(self, foo, bar):
        return InputObject(foo, bar)
```

```rust
let input = model.method_is("make_input_object", &[
    &tch::Tensor::of_slice(&[1_f32, 2_f32, 3_f32]).into(),
    &tch::Tensor::of_slice(&[4_f32, 5_f32, 6_f32]).into(),
]);
let output = model.method_is("add_them", &[&input.unwrap()]);
```

LaurentMazare:
I just pushed a fix that should hopefully get rid of this and be more principled than #476 . This handling of jit Object is certainly a very tricky bit though so let me know if you see further issues.

PR #476, opened 2022-04-21
Avoid double-free of borrowed Object

neilisaac:
resolves https://github.com/LaurentMazare/tch-rs/issues/475

### DIFF
--- a/src/wrappers/jit.rs
+++ b/src/wrappers/jit.rs
@@ -250,16 +250,19 @@ impl IValue {
                 let v = v.iter().map(Self::to_c).collect::<Result<Vec<_>, TchError>>()?;
                 let tuple = ati_tuple(v.as_ptr(), v.len() as c_int);
                 for x in v {
-                    ati_free(x);
+                    if unsafe_torch_err!(ati_tag(x)) != 14 {
+                        unsafe { ati_free(x) };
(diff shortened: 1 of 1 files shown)


LaurentMazare:
Thanks for the PR, maybe you could add some comments to give some insights on why it's safe to handle tag 14 differently from the others?

neilisaac:
@LaurentMazare I was hoping you'd have more context.  I'm not an expert on libtorch internals.  

I see that Object holds a pointer which it frees on drop: https://github.com/LaurentMazare/tch-rs/blob/ea61a05ce2190c422e41d208422e3ba60dfad078/src/wrappers/jit.rs#L752 and came to the notion that it may be a double-free issue based how Object is handled here: https://github.com/LaurentMazare/tch-rs/blob/ea61a05ce2190c422e41d208422e3ba60dfad078/src/wrappers/jit.rs#L309

I only tested the case for Object as an argument to CModule.method_is: https://github.com/LaurentMazare/tch-rs/issues/475

neilisaac:
https://github.com/LaurentMazare/tch-rs/commit/2d7c2cf868c76aba7ffd18f4d7316716815c4041 looks much nicer, thanks @LaurentMazare!

Issue #437, opened 2021-12-18
How to implement torch.autograd.Function in rust?


any35:
Hi, all.

I try to rewrite a python lib in rust. But I have some problem when I try to implement a custom backward method.

there is the sample code

```python
    class MyMethod(torch.autograd.Function):
        @staticmethod
        def forward(ctx, var1, var2):
            # sth....
            return ans

        @staticmethod
        def backward(ctx, *grad_ans):
             # sth....
            return (var1_grad, var2_grad)

```

Is there any equivement impl  in rust?

Thank you.

LaurentMazare:
Hello,
Sadly there is no good way to implement a custom backward method at the moment. This was discussed in #279 and #70 I think. A simple version of this should not be very hard to implement but nobody has got into it yet.

any35:
Thank you for your reply. 
I will also look into it to see if I can solve this problem.

PR #478, opened 2022-04-23
add custom function, part 1

any35:
hi,  @LaurentMazare 

I opened an issue last year, and now I finally have time to implement it. 
#437  

This is the first part and the easiest way to implement.  

For details, you can see the test code.

This is just a workaround, but useful in many cases.

I'm going to implement the real "custom function" part next week.

Thank you for your review.


duanckham:
@any35 
How it going, why you close it, bro?

PR #485, opened 2022-05-16
expose `getattr` for jit object

0dlinwaabi:
## Summary
- expose `getattr` for `tch::jit::object`
  - allow access to the torch jit object attributes

### DIFF
--- a/torch-sys/libtch/torch_api.cpp
+++ b/torch-sys/libtch/torch_api.cpp
@@ -1408,11 +1408,18 @@ ivalue ati_object_method_(ivalue i, char *method_name, ivalue *ivalues, int niva
   return nullptr;
 }
 
+ivalue ati_object_getattr_(ivalue i, char *attr_name) {
+  PROTECT(
+    torch::jit::IValue output  = i->toObjectRef().getAttr(attr_name);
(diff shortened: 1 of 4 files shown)


torch-sys/libtch/torch_api.cpp
```diff
@@ -1408,11 +1408,18 @@ ivalue ati_object_method_(ivalue i, char *method_name, ivalue *ivalues, int niva
...
+    return new torch::jit::IValue(output);
+  )
+  return nullptr;
+}
+
 ivalue ati_clone(ivalue i) {
   PROTECT(
     return new torch::jit::IValue(*i);
   )
-  return nullptr;
```
LaurentMazare, line 1415: Why was this removed? Could you put it back as it's useful to avoid warnings for some compilers?

LaurentMazare:
There are also some rustfmt and clippy failures in the CI, could you try running both?

LaurentMazare:
It seems that rustfmt is still broken in the CI.

LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR.

PR #486, opened 2022-05-18
Orthogonal weight initialization, RNN weight and bias config

bripeace:
This PR adds Orthogonal Matrix weight initialization to nn::Init and exposes weight and bias initialization in RNNConfig.

Orthogonal matrix initialization is useful in reinforcement learning and used extensively in policy-based methods. The init code is based on the [pytorch code](https://pytorch.org/docs/stable/_modules/torch/nn/init.html#orthogonal_). There is a [c++ api function](https://pytorch.org/cppdocs/api/function_namespacetorch_1_1nn_1_1init_1a5978fcc257460475f635b5960e892a8e.html#_CPPv4N5torch2nn4init11orthogonal_E6Tensord), but it wasn't clear to me how to expose those functions for use. If it's preferable to use the c++ api point me to a good example of how to expose them and I can adjust the code accordingly.
Added two tests, one to check the init function works properly and second to ensure the check for shape requirements is present.

The RNNConfig exposes weight and bias initialization like is done for Linear and Embedding modules. I also factored out the for loops to its own function since they were practically identical for GRU and LSTM. I did not add a test to check that the proper number of layers were created as flat_weights is marked private and wasn't sure about the reasoning there. If there's a better way to test that or if your open to making flat_weights public I can add them.

I did run into something strange when writing the orthogonal test: Tensor::norm does not return consistent values. It's either a little high or a little low vs the expected value Tensor::linalg_norm gives. Torch::norm is deprecated so maybe this isn't a concern, but the python version is exact every run.

### DIFF
--- a/src/nn/var_store.rs
+++ b/src/nn/var_store.rs
@@ -530,6 +530,20 @@ impl<'a> Path<'a> {
         self.f_var(name, dims, Init::KaimingUniform)
     }
 
+    /// Creats a new variable initialized randomly with an orthogonal matrix
+    ///
+    /// The new variable is named according to the name parameter and
+    /// has the specified shape. The variable is trainable, its gradient
+    /// will be tracked.
+    /// The variable uses a float tensor initialized randomly with an orthogonal
+    /// matrix as described in *Exact solutions to the nonlinear dynamics
+    /// of learning in deep linear neural networks* - Saxe, A. et. al. (2013).
+    /// The input tensor must have at least 2 dimensions, and for tensors
+    /// with more than 2 dimensions the trailing dimensions are flattened.
+    pub fn f_orthogonal(&self, name: &str, dims: &[i64], gain: f64) -> Result<Tensor, TchError> {
+        self.f_var(name, dims, Init::Orthogonal { gain })
+    }
+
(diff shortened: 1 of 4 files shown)


src/nn/var_store.rs
```diff
@@ -530,6 +530,20 @@ impl<'a> Path<'a> {
         self.f_var(name, dims, Init::KaimingUniform)
     }
 
+    /// Creats a new variable initialized randomly with an orthogonal matrix
```
LaurentMazare, line 533: Could you fix the typo `Creats` -> `Creates`.

bripeace: Corrected. Thanks.

src/nn/var_store.rs
```diff
@@ -638,6 +652,20 @@ impl<'a> Path<'a> {
         self.f_kaiming_uniform(name, dims).unwrap()
     }
 
+    /// Creats a new variable initialized randomly with an orthogonal matrix
```
LaurentMazare, line 655: Same typo as above?

LaurentMazare:
Merged, thanks for the PR!

PR #495, opened 2022-06-08
Update time dependency

fredr:
`zip 0.5.x` depends on an old version of the time crate affected by this https://rustsec.org/advisories/RUSTSEC-2020-0071
It was updated to a newer version in `zip 0.6.x`


LaurentMazare:
Merged, thanks for the PR!

Issue #494, opened 2022-06-07
Addition of VarStore merging capabilities


guillaume-be:
Hello,

As per https://github.com/guillaume-be/rust-bert/issues/258, I am unsure if it is currently possible to load multiple pre-trained models into a given `VarStore`. A possibility would be for the user to add unique prefixes for each of the models pretrained weights and then use consecutive `load_partial` calls. This has the drawback of losing the validation or the weights names performed during a standard `load` in a `VarStore`.

I have created a draft implementation allowing to take multiple `VarStore` and merging them into a single one for further training by passing them to an optimizer. This allows the following:

```rust
let vs_1 = VarStore::new(Device::Cpu);
let _ = vs_1.root().entry("key_1").or_zeros(&[3, 1, 4]);
let _ = vs_1.root().entry("key_2").or_zeros(&[1, 5, 9]);

let vs_2 = VarStore::new(Device::Cpu);
let _ = vs_2.root().entry("key_3").or_zeros(&[2, 4, 4]);
let _ = vs_2.root().entry("key_4").or_zeros(&[5, 2, 3]);

let merged_vs = VarStore::from_existing(vec![vs_1, vs_2]);
```

The method would look as follows: https://github.com/LaurentMazare/tch-rs/compare/main...guillaume-be:master

If you are supportive of the idea I'd be happy to open a PR to bring these changes to the main branch. Please let me know if there would be another preferred way to load multiple models into a single `VarStore` for combine optimization.

Thank you!

LaurentMazare:
Hi,
This sounds reasonable to me, I indeed don't think that there is a very good way to do this at the moment.
If the main goal is passing all weights to an optimizer, an alternative could be to tweak the optimizer side to allow for this, but maybe there is also some upside in being able to store all the combined variables in a single file in the end.
Some nitpicking:
- Re naming, maybe `merge` or `combine` instead of `from_existing`?
- Instead of taking a slice that must have the same length as the var-store vec, it might be better to take a `Vec<(VarStore, Option<&str>)>` or something like this?

lumost:
> If the main goal is passing all weights to an optimizer, an alternative could be to tweak the optimizer side to allow for this, but maybe **there is also some upside in being able to store all the combined variables in a single file in the end.**

This would indeed be beneficial.  The use case that prompted this was related to combining two pre-trained models into a new model, training on a specific task, and then ideally writing the new model out.  If the new model could be written and loaded as a single artifact , it would provide a nice ease of use win.

PR #496, opened 2022-06-08
Addition of VarStore merging capabilities 

guillaume-be:
Addition of merge capabilities for `VarStore`:
- an input `Vec` of `VarStore` is consumed and the variables are moved into a new combined `VarStore`
- optional prefixes can be provided and prepended to the `VarStore` variables, for example to avoid key conflicts.

Fixes https://github.com/LaurentMazare/tch-rs/issues/494

### DIFF
--- a/src/nn/var_store.rs
+++ b/src/nn/var_store.rs
@@ -59,29 +59,17 @@ impl VarStore {
         VarStore { variables_: Arc::new(Mutex::new(variables)), device }
     }
 
-    pub fn from_existing(
-        mut var_stores: Vec<VarStore>,
-        prefixes: Option<&[&str]>,
-    ) -> Result<VarStore, TchError> {
+    pub fn merge(var_stores: Vec<(VarStore, Option<&str>)>) -> Result<VarStore, TchError> {
         let mut new_var_store = VarStore::new(Device::Cpu);
 
         if var_stores.is_empty() {
             Ok(new_var_store)
-        } else if var_stores.len() == 1 {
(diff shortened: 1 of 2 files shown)


src/nn/var_store.rs
```diff
@@ -59,6 +60,55 @@ impl VarStore {
...
+                                "Duplicate variable name found: {}. Provide a unique prefix to allow merge operation",
+                                v.key(),
+                            )));
+                        }
+                        Vacant(v) => {
+                            v.insert(var);
+                        }
+                    }
+                }
+                for trainable_var in Arc::try_unwrap(var_store.variables_)
```
LaurentMazare, line 95: What guarantees do we have that this would not panic? Maybe cloning the variables would be safer if this makes sense?

guillaume-be: Thank you this is a good point. I am unsure about the implication of cloning variables (do the new `VarStore` variables and tensor still correspond to the variables that may be loaded into a model in the input varstores?). I have instead updated it and not assume no other thread has a copy of the `Arc`, but rather lock it and drain the underlying variables vector. This means the older `VarStore` will be invalid (empty variables) but this was the intention of this method in any case.

LaurentMazare (approved):

src/nn/var_store.rs
LaurentMazare: Right, as it's passing the argument `VarStore` as values and so consuming them, this seem reasonable.

PR #500, opened 2022-06-22
Add more function from torch::cuda

vcarpani:
This PR adds some more torch::cuda functions and improves documentation for the existing ones.


vcarpani:
Hello, any updates on this? 

In addition to that, my company is really interested in supporting this project since we are using it in production, can we help in any way with the maintenance of this repo?

src/wrappers/device.rs
```diff
@@ -12,32 +12,59 @@ pub enum Device {
 /// Cuda related helper functions.
 pub enum Cuda {}
 impl Cuda {
-    /// Returns the number of GPU that can be used.
+    /// Returns the number of CUDA enabled GPUs available.
```
LaurentMazare, line 15: Maybe we should re-use the same help text as what appears on the C++ side, e.g. [here](https://github.com/pytorch/pytorch/blob/master/torch/csrc/api/include/torch/cuda.h#L11)?

Also having the documentation duplicated 3 times doesn't sound very nice so I would be in favor of not having it for the C++ side and `torch-sys` (these are only used as internal dependencies anyway).

torch-sys/libtch/torch_api.cpp
```diff
@@ -925,6 +925,18 @@ int atc_cudnn_is_available() {
   return -1;
 }
 
+void atc_manual_seed(int64_t seed) {
```
LaurentMazare, line 928: Shouldn't this be an uint rather than an int?

LaurentMazare:
Sorry for the delay, I've added some comments. It's great to know that this is used in production, it would probably be helpful to know a bit more on how it's used / if there are some repos based on it so as to involve you e.g. when potentially introducing breaking changes.

vcarpani:
> Sorry for the delay, I've added some comments. It's great to know that this is used in production, it would probably be helpful to know a bit more on how it's used / if there are some repos based on it so as to involve you e.g. when potentially introducing breaking changes.

No problems about the delay, thanks for maintaining this repo. We are currently using it in 2 different repos, one using the JIT loader and the other one using torch operators directly. 

For breaking changes, I would propose to use a `CHANGELOG` file, if you believe it is a good idea, I can open a PR on the CHANGELOG format, so that all the people using the repo would have an idea of what changed and which version to be used in their productions environment

src/wrappers/device.rs
vcarpani: I agree on using the official documentation, I will solve that.

It does not look nice, but for me, it makes it easier to maintain the code when more people are collaborating on it

torch-sys/libtch/torch_api.cpp
vcarpani: Yes, will fix that

src/wrappers/device.rs
LaurentMazare: Just to mention that I'm waiting on this one to be fixed before merging the PR, the rest looks all good to me.

vcarpani: Should be solved now, I also updated it to use the new changelog

CHANGELOG.md
```diff
@@ -3,7 +3,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## v0.10.0 - not released yet
+## Unreleased
```
LaurentMazare, line 6: Could you revert this line to mentioning v0.10.0 as it's the version in the `Cargo.toml` on the github tip.

CHANGELOG.md
```diff
@@ -3,7 +3,11 @@ All notable changes to this project will be documented in this file.
...
+### Added
+- Expose functions for setting manual seeds for CUDA devices.
```
LaurentMazare, line 8: Add a link to the PR?

vcarpani:
Should be solved now ;)

LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR!

PR #508, opened 2022-07-05
Supporting cu116 at build time.

Narsil:
Currently the build fails because of some import on 
libtorch==1.12.0. 

This PR takes care of `cu116`.


LaurentMazare:
Merged, thanks for the PR!

PR #514, opened 2022-07-28
Build script fixed: includes separated from symbols.

npatsakula:
Hello! I found little bug in `build.rs` and would like to help :D

## Problem

NixOS distributive (and some more) separating `libtorch` shared symbols and headers. And it would be great if we add ability to separate ENV variables for them.

## Solution

- [x] Add ENV variable (keep reverse compatibility). 
- [x] Add documentation.
- [x] Make extra test.

### DIFF
--- a/torch-sys/build.rs
+++ b/torch-sys/build.rs
@@ -147,7 +147,8 @@ fn prepare_libtorch_dir() -> PathBuf {
 
 fn make<P: AsRef<Path>>(libtorch: P, use_cuda: bool, use_hip: bool) {
     let os = env::var("CARGO_CFG_TARGET_OS").expect("Unable to get TARGET_OS");
-    let includes: PathBuf = env::var("LIBTORCH_INCULDES").map(Into::into).unwrap_or_else(|_| libtorch.as_ref().to_owned());
+    let includes: PathBuf = env_var_rerun("LIBTORCH_INCLUDE").map(Into::into).unwrap_or_else(|_| libtorch.as_ref().to_owned());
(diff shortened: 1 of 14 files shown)


torch-sys/build.rs
```diff
@@ -168,8 +169,8 @@ fn make<P: AsRef<Path>>(libtorch: P, use_cuda: bool, use_hip: bool) {
                 .cpp(true)
                 .pic(true)
                 .warnings(false)
-                .include(libtorch.as_ref().join("include"))
-                .include(libtorch.as_ref().join("include/torch/csrc/api/include"))
+                .include(&includes)
```
LaurentMazare, line 172: I don't feel that this would work in the standard setting as we need the `include` subdirectory of `libtorch` but maybe I'm missing something here?

npatsakula: Thank you for reply.

Yes, you're right, I'll fix it in a moment.

npatsakula:
@LaurentMazare, hello! Can i ask for review?

LaurentMazare (approved):

LaurentMazare:
Looks reasonable, could you rebase this branch? Hopefully this should get rid of the clippy error in the CI.

npatsakula:
> Looks reasonable, could you rebase this branch? Hopefully this should get rid of the clippy error in the CI.

Hello @LaurentMazare! I rebased my branch on new `main`, added small documentation paragraph into `README.md` and performed small improvements in `build.rs` code.

Now I think this MR is ready to merge.

LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR.

PR #522, opened 2022-08-19
Added missing InstanceNorm module.

joverwey:
I noticed that InstanceNorm was not implemented. I tried to follow the same pattern as BatchNorm with a few exceptions.
For BatchNorm, the current implementation always uses weights and biases and always keeps track of running stats.
In the PyTorch implementation this is optional. For InstanceNorm I do need to be able to enable/disable this so I added flags for
- affine
- track_running_stats
 
like the PyTorch implementation has.

Let me know if I missed anything.

### DIFF
--- a/src/nn/instance_norm.rs
+++ b/src/nn/instance_norm.rs
@@ -1,0 +1,135 @@
+//! An instance-normalization layer.
+use crate::Tensor;
+
+use std::borrow::Borrow;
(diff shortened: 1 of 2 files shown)


joverwey:
Clippy seems to be failing in a file I didn't touch? Any idea how I can fix this?

PR #562, opened 2022-11-06
Add Device::Mps

danieldk:
This adds `Device::Mps`, which is necessary to run kernels using Metal Performance Shaders.

I confirmed that I can run SyntaxDot models on the GPU. Unfortunately, it is still much slower than the `Cpu` device that uses the AMX co-processor on M1 Macs. SyntaxDot is a fairly large codebase/set of models and I seem to be hitting some degenerate cases that I don't have time to debug now, but at least the device support seems to work.

### DIFF
--- a/src/wrappers/device.rs
+++ b/src/wrappers/device.rs
@@ -7,6 +7,8 @@ pub enum Device {
     Cpu,
     /// The main GPU device.
     Cuda(usize),
+    /// The main MPS device.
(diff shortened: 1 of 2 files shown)


LaurentMazare (approved):

LaurentMazare:
Looks good, thanks for the PR, hopefully this will get faster than cpu for your use case soon.

PR #572, opened 2022-11-18
Iter2: store my output on the same device as the first input

njaard:
In order for `Iter2` to be usable with a GPU, you needed to do `Iter2::new(...).to_device(vs.device()).shuffle()`

Now with this patch, the `to_device` is not necessary, which is consistent with my expectations.

However, it has virtually no performance difference, positive or negative.


vcarpani:
Thanks for this PR, what happens if `xs` and `ys` are on different devices?

njaard:
@vcarpani It doesn't work, which is "philosophically" the same behavior as right now, but incompatible, because right now if the first parameter is GPU and the second is CPU, then Iter2 will work. With my patch. The first parameter must be the same as the second or you'll need to move the second it to the GPU.

Issue #563, opened 2022-11-10
Unresolved external symbols related to TLS


dechenio:
I have been struggling to compile the example using prebuilt libtorch+cu117 for Windows (11) with Rust (1.64)-msvc (VS2022). I can build the tch-rs library without error but when linking to any application either my own simple code or the bundled example (i.e. basics), the following errors appeared: 

```shell
$$> $env:TORCH_CUDA_VERSION = "cu117"
$$> $env:LIBTORCH = "c:\dev\libtorch\release"
$$> $env:Path += ";C:\dev\libtorch\release\lib"
$$> $env:RUSTFLAGS = "-L c:\dev\libtorch\release\lib -l c10_cuda"
$$> cargo run --example basics --release
..........
error: linking with `link.exe` failed: exit code: 1120
........
 Creating library D:\projects\rust\tch-rs\target\release\examples\basics.lib and object D:\projects\rust\tch-rs\target\release\examples\basics.exp
          libtorch_sys-18bee47690c056b1.rlib(torch_api.o) : error LNK2001: unresolved external symbol __imp___tls_index_?init@?1??lazy_init_num_threads@internal@at@@YAXXZ@4_NA
          libtorch_sys-18bee47690c056b1.rlib(torch_api.o) : error LNK2001: unresolved external symbol __imp___tls_offset_?init@?1??lazy_init_num_threads@internal@at@@YAXXZ@4_NA
          D:\projects\rust\tch-rs\target\release\examples\basics.exe : fatal error LNK1120: 2 unresolved externals
```

It seems something wrong with linking thread local storage related symbols.  Would you help us to fix it? Many thanks.

LaurentMazare:
I don't know much about windows linking and it seems that it's likely to be an issue with the underlying tls crate rather than with `tch`. That said, afaik `tch` only uses tls through the `curl` dependency which is used to fetch the libtorch binaries automatically. If you have already downloaded libtorch and set the `LIBTORCH` environment variable as seems to be the case you could try disabling this feature. This should hopefully get rid of the `curl` dependency and of tls.
In your `Cargo.toml`, you will need something like:
```
tch = { version = "0.9.0", default-features = false }
```

dechenio:
Thanks @LaurentMazare .
I have tried as above but the same errors related to tls. 
[...]

PR #580, opened 2022-12-03
Fix Windows CI checks

guillaume-be:
Hello,

The latest version of Visual Studio (17.4) introduces a regression and causes issue for the compilation/linking of Pytorch (see https://developercommunity.visualstudio.com/t/thread_local-causing-fatal-error-LNK1161/10199441). Unfortunately this affects the github actions relying on `windows-latest` as they automatically include VS17.4.

Since the issue does not occur for earlier versions (17.3 and lower), downgrading temporarily the windows server version to 2019 solves the issue and allows building the library again. I have made the changes on the [rust-bert](https://github.com/guillaume-be/rust-bert/pull/304) repository and confirms this allows running the test on a windows machine as well.

I am opening this PR if you would be interested in fixing the Windows CI while the Microsoft team works on a fix. I will periodically check for a resolution of the issue on `windows-latest` on the rust-bert repo, I'd be happy to let you know/open a PR to upgrade again when the issue is fixed.

Related to https://github.com/LaurentMazare/tch-rs/issues/563

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macOS-latest]
+        os: [ubuntu-latest, windows-2019, macOS-latest]
         rust: [stable, nightly]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


LaurentMazare (approved):

LaurentMazare:
Great, thanks for the fix!

Issue #145, opened 2020-01-22
Does Adam optimizer missing eps field?


lykhouzov:
Hello.
When I tried to play with AdamConfig params, I found that there is no `eps` option, where in official documentation of pytorch it exists
```
torch.optim.Adam(params, lr=0.001, betas=(0.9, 0.999), eps=1e-08, weight_decay=0, amsgrad=False)
```
Then I also checked in C++ Api and I am not sure, because I am not familiar with C++, but there is a
```
auto eps(const double &new_eps)
```
in `AdamOption` struct, which force me to think that there is a `eps` option for Adam optimizer..

So the question is, does rust implementation missing this param or it uses standard `1e-8` some where?

Thank you.

PS: `tch="0.1.3"`

LaurentMazare:
Indeed `AdamOption` does not include this `eps` field. Calls to the adam optimizer in rust end up calling the C++ api via [ato_adam](https://github.com/LaurentMazare/tch-rs/blob/bfe0ae0962c5c9c3d607734325c8e7ab99b6329d/torch-sys/libtch/torch_api.cpp#L377-L390). This uses the default value for `eps` from the C++ implementation which is 1e-8 as you can see [here](https://github.com/pytorch/pytorch/blob/master/torch/csrc/api/include/torch/optim/adam.h#L27).

lykhouzov:
Thank you, @LaurentMazare 
I couldn't find this.
I'm closing the ticket.

LaurentMazare:
If exposing eps to rust would be useful to you, feel free to make a PR for it (or just let me know and I'll do it at some point).

PR #600, opened 2022-12-31
Add eps and amsgrad options to Adam Optimizer

zeyugao:
Expose more options in AdamOptions

Fix https://github.com/LaurentMazare/tch-rs/issues/145

However it will break the `adamw` and `adam` interface under `src/nn/optimizer.rs`. Should I create a new functions like `adam_new`?

### DIFF
--- a/examples/gan/main.rs
+++ b/examples/gan/main.rs
@@ -96,11 +96,11 @@ pub fn main() -> Result<()> {
 
     let mut generator_vs = nn::VarStore::new(device);
     let generator = generator(generator_vs.root());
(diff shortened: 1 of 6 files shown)


LaurentMazare:
Thanks for the PR, I've made further changes to preserve the current api and the newly added field can still be configured via a builder pattern. I think I'll lilkely remove the `nn::adam` and `nn::adamw` functions altogether as the intended way to setup the optimizer would be as follows, though I want to think a bit more about it before removing these.
```rust
let opt = Adam::default().eps(1e-6).amsgrad(true).build_copt(...)
```

Issue #609, opened 2023-01-15
Torch("supported devices include CPU, CUDA and HPU, however got MPS")


LeeeSe:
torch-version: 1.13.1
OS: macOS 13.1
Chip: M1

```
use tch::{kind, Tensor};

fn main() {
    let model =
        tch::CModule::load_on_device("model/yolov5s.torchscript", tch::Device::Mps).unwrap();
}
```

LaurentMazare:
I think the issue is actually on the libtorch side, see https://github.com/pytorch/pytorch/issues/88820
A workaround seems to be to load the model to the cpu and then move it to mps.

LeeeSe:
Thanks, the problem was solved by the following
```
fn load_model_to_gpu() {
    let mut vs = tch::nn::VarStore::new(tch::Device::Cpu);
    vs.load("model/yolov5s.torchscript").unwrap();
    println!("vs.device {:?}", vs.device());
    vs.set_device(tch::Device::Mps);
    println!("vs.device now {:?}", vs.device());
}
```

guillaume-be:
@LaurentMazare the workaround seems to work for users with access to the `VarStore`, but causes some frictions when the application does not expose it anymore (e.g. higher level model APIs), see https://github.com/guillaume-be/rust-bert/issues/311

Since this seems to be the only way to load weights on MPS currently, would you be open to a pull request making this the default behaviour when loading weights (i,e. if the `VarStore` device is `Device::Mps`, change it to `Device::Cpu`, load the weights and then change back to `Device::Mps`)? 

This hopefully gets resolved eventually and the workaround could then be removed from the `load` method.

LaurentMazare:
@guillaume-be sure, that sounds like a reasonable temporary workaround so feel free to mint a PR or let me know if you don't have the time and I'll do it, probably good to have a comment that points at this issue and mention that we should revisit.

PR #623, opened 2023-02-12
MPS Load capabilities

guillaume-be:
- Temporary fix to the `VarStore` load method allowing loading weights on MPS-accelerated machines
The device of the `VarStore` is temporarily set to `CPU` when loading the weights and then changed back to `MPS` as described by the workarounds described in https://github.com/guillaume-be/rust-bert/issues/311 and https://github.com/LaurentMazare/tch-rs/issues/609

I do not have a `MPS` device to test the behaviour. It would be great if one of the contributors who identified the issue and the workaround could validate it solves the loading issues before merging.


bakkot:
I am not one of the original contributors, but I do have an MPS device, and I confirm this fixed the issue.

Specifically, I built https://github.com/LaurentMazare/diffusers-rs with the device manually set to MPS by changing [this line](https://github.com/LaurentMazare/diffusers-rs/blob/44bb82f76c3deab5a1d26854f4ac9a01003bed2c/examples/stable-diffusion/main.rs#L233) to `Device::Mps`. When specifying this crate as a dependency at `main`, that fails with the error mentioned in https://github.com/LaurentMazare/tch-rs/issues/609. When I use this crate at this PR, it works, and runs several times faster than the non-MPS version.

While I'm here, I note that this PR doesn't restore the original device in the two error cases (the `?` and the explicit `return Err`). That seems like it might be confusing.

LaurentMazare:
Thanks for confirming, agreed that it would be nicer to set the device back to `Mps` if an error causes an early exit, @guillaume-be is it something that you could look at and then I think this is good to be merged.

guillaume-be:
Thank you very much for confirming @bakkot - @LaurentMazare yes I should be able to implement a more robust setup later today

guillaume-be:
@bakkot , @LaurentMazare I have wrapped the tensor loading in a closure to ensure the device reset code gets executed before the closure `Result` gets passed on to the caller

LaurentMazare (approved):

LaurentMazare:

PR #624, opened 2023-02-16
Updated zero_state method of RNNs to use Double type Tensor if needed.

KapBenjamin:
zero_state method of LSTM and GRU now retrieve 'Kind' from the first Tensor of flat_weights. Previously was hardcoded as 'Kind::Float', which caused dtype mismatch error when weights were of Kind Double at that point.


LaurentMazare (approved):

LaurentMazare:

PR #635, opened 2023-03-07
Add rust Image support

oovm:
Added From trait to support rust's `image` ecology

### DIFF
--- a/src/vision/rust_image.rs
+++ b/src/vision/rust_image.rs
@@ -8,19 +8,21 @@ use image::{
 use crate::vision::image::{chw_to_hwc, hwc_to_chw};
 use crate::{Kind, TchError, Tensor};
 
-impl<'i> From<&'i DynamicImage> for Tensor {
-    fn from(image: &'i DynamicImage) -> Self {
+impl<'i> TryFrom<&'i DynamicImage> for Tensor {
+    type Error = TchError;
+
+    fn try_from(image: &'i DynamicImage) -> Result<Self, Self::Error> {
         match image {
-            DynamicImage::ImageLuma8(gray) => Tensor::from(gray),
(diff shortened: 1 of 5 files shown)


src/vision/rust_image.rs
```diff
@@ -0,0 +1,180 @@
...
+
+use image::{
+    DynamicImage, EncodableLayout, GrayAlphaImage, GrayImage, Rgb32FImage, RgbImage, Rgba32FImage,
+    RgbaImage,
+};
+
+use crate::vision::image::{chw_to_hwc, hwc_to_chw};
+use crate::{Kind, TchError, Tensor};
+
+impl<'i> From<&'i DynamicImage> for Tensor {
```
LaurentMazare, line 11: As this is fallible, it might be better to use try-from and let the end user unwrap and handle the error if they want. The same applies to the other `From` impl below.

src/vision/rust_image.rs
```diff
@@ -0,0 +1,180 @@
...
+            "Tensor kind is `{kind:?}`, except `{except:?}` tensor",
+            kind = kind,
+            except = except
+        );
+        Err(TchError::Convert(msg))?;
+    }
+    Ok(chw_to_hwc(tensor))
+}
+
+// #[test]
```
LaurentMazare, line 158: Could you either enable or remove these tests?

oovm: There is no suitable image in `tests/`, if there is, I can move to the `tests/` directory.

src/vision/rust_image.rs
oovm: Ok, now all methods use `f_*` version and raise error of `TchError`

LaurentMazare (approved):

LaurentMazare:
Merged, thanks for the PR!

PR #633, opened 2023-03-05
Support load data from `*.safetensors`

oovm:
Fix #575

---

Hi @LaurentMazare , can you take a moment of time to this, I need this feature for debugging on some models.

PR #636, opened 2023-03-08
Support safetensors format

A2va:
This is a mostly the same as #633, but there are some key differences:

* The safetensors version is newer this PR. This allow to remove an unsafe part.
* Some unit tests were made.
* Add `is_contiguous` function

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ Cargo.lock
 __pycache__
 *~
 .*~
+.vscode/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ thiserror = "1"
 torch-sys = { version = "0.10.0", path = "torch-sys" }
 zip = "0.6"
 half = "1.6"
+safetensors = "0.2.8"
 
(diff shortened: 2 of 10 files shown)


oovm:
@A2va , Can you add `fill/read` api for `VarStore`?

- read: With VarStore as the dominant, VarStore will report an error if the variable cannot be found in safetensors
- fill: With safetensors as the dominant, redundant variables in safetensors will be skipped.

Of course you can choose other names, but these two interfaces represent different usage situations.

oovm:
And I think all api should accept `&[u8]` instead of path, which can be load via mmap if it's on disk, or slice if in memory.

A2va:
> And I think all api should accept `&[u8]` instead of path, which can be load via mmap if it's on disk, or slice if in memory.

I don't know how to feel about this, all the I/O function already  used Path. I think it will be inconsistent to have safetensors with `&[u8]` and not the others. Maybe @LaurentMazare has his own opinion on this?

A2va:
@oovm I have added what you ask for.

LaurentMazare (approved):

Issue #645, opened 2023-03-14
Element implementation for `f16` / `bf16`


guillaume-be:
Hello,

I am trying to copy the data from a `Kind::Half` Tensor to a `Vec<f16>` and I am having issues using the automated conversion utilities (the same happens for `BFloat16`).
The following code fails with `the trait Element is not implemented for f16`:
```rust
use tch::{Device, Kind, Tensor};
fn main() {
    let tensor = Tensor::zeros(&[42], (Kind::Half, Device::Cpu));
    let vector: Vec<half::f16> = tensor.into();
}
```
while this runs without issues:
```rust
fn main() {
    let tensor = Tensor::zeros(&[42], (Kind::Float, Device::Cpu));
    let vector: Vec<f32> = tensor.into();
}
```

I tried re-implementing the method at a lower level without success (same error):
```rust
let numel = tch_tensor.numel();
let mut vec = vec![half::f16::ZERO; numel];
tch_tensor.to_device(Device::Cpu).to_kind(Kind::Half).copy_data(&mut vec, numel);
```

I am a bit confused since `Element` seems to be implemented for both these types https://github.com/LaurentMazare/tch-rs/blob/ac0135f822d9ba27bf002f984e6523690af61b38/src/wrappers/kind.rs#L136

Do you have an idea on how I could perform such a conversion?
Thank you!

LaurentMazare:
That's odd, I actually just gave this a try with the latest github tip version and it worked out of the box for me. Are you using the latest version?
```bash
$ cargo run --example basics 
     Running `target/debug/examples/basics`
 0
 0
 0
[ CPUHalfType{3} ]
[0.0, 0.0, 0.0]
$ cat examples/basics.rs 
use tch::{Device, Kind, Tensor};

fn main() {
    tch::maybe_init_cuda();
    let tensor = Tensor::zeros(&[3], (Kind::Half, Device::Cpu));
    tensor.print();
    let vector: Vec<half::f16> = tensor.into();
    println!("{:?}", vector);
}
```

guillaume-be:
Indeed - I just checked and I am using `half=2` as a dependency in my project while `tch` depends on `1.6`. Because of the major version difference, the implementation does not apply to the `f16` of the `v2` crate.

Thank you - I believe I will roll back to the previous `half` version  to `1.x` for now and this solves the issue

LaurentMazare:
[...]

PR #646, opened 2023-03-14
`half` version update

guillaume-be:
Hello,
With reference to https://github.com/LaurentMazare/tch-rs/issues/645, this small PR proposes an update of the version of `half`. Unfortunately rolling back to `1.x` is challenging as other dependencies (including https://crates.io/crates/ort) start upgrading to `2.x`. 

Note that this would be a breaking change for crates that have other dependencies that depend on `half` (as illustrated by https://github.com/LaurentMazare/tch-rs/issues/645). This is relevant because it seems `1.x` is still more popular than `2.x`: https://crates.io/crates/half .


LaurentMazare:
Thanks, using the latest stable version sounds reasonable so let's move to 2 and I'll make sure to only release this as part of a major version change.

LaurentMazare (approved):

PR #648, opened 2023-03-15
impl Element for bf16

guillaume-be:
Hello,

Relates to https://github.com/LaurentMazare/tch-rs/issues/645, I noticed that while `f16` implements `Element`, `bf16` doesn't. This small PR proposes to add support for the same.
Thank you!


LaurentMazare (approved):

LaurentMazare:
Looks like a good thing to have, thanks for the PR!

PR #651, opened 2023-03-18
Update for libtorch 2.0.0

LaurentMazare:


gen/gen.ml
simlay, line 1: I spent a bit of time on this 4 days ago but gave up after the generated code didn't work. The fact that this required modifying the ocaml generation part makes me feel less inadequate.

PR #661, opened 2023-03-27
Use PyPi API to get url for the prebuilt macOS M1 python wheel

### LINKED ISSUE #629: tch-rs does not run on m1 mac
`tch-rs` does not run on m1 MacOS. To fix the problem I followed these steps adopted from https://github.com/ssoudan/tch-m1.

1. pip3 install torch
2. export paths:
```bash
export LIBTORCH=$(python3 -c 'import torch; from pathlib import Path; print(Path(torch.__file__).parent)')
export DYLD_LIBRARY_PATH=${LIBTORCH}/lib

```

simlay:
I was curious how python wheels were compressed/bundled. Turns out it's a [zip file](https://peps.python.org/pep-0427/#abstract) so I figured I'd show the viability of using the PyPi API to get the build for arm64 macOS.

I'm not in love with it as it's kind of a hack. Here are the two issues:
* The wheel unzips to `torch` rather than `libtorch` so the extract function has to handle this.
* The `find_map` string, `torch-{TORCH_VERSION}-cp311-none-macosx_11_0_arm64.whl` is likely to change a bit in future releases. I was tempted to use `filename.contains("arm64") && filename.contains("macosx")` but that is still [8 different wheels (python 3.8 to 3.11 on macOS 11 and macOS 10.9)](https://pypi.org/project/torch/2.0.0/#files).

In theory, this closes #629.

Let me know if this seems like a reasonable path forward.

### DIFF
--- a/torch-sys/build.rs
+++ b/torch-sys/build.rs
@@ -32,6 +32,46 @@ fn download<P: AsRef<Path>>(_source_url: &str, _target_file: P) -> anyhow::Resul
     anyhow::bail!("cannot use download without the ureq feature")
 }
 
+#[cfg(not(feature = "download-libtorch"))]
+fn get_pypi_wheel_url_for_aarch64_macosx() -> anyhow::Result<&str> {
+    anyhow::bail!("cannot get pypi wheel url without the ureq feature")
+}
+
+#[cfg(feature = "download-libtorch")]
+#[derive(serde::Deserialize, Debug)]
+struct PyPiPackageUrl {
+    url: String,
+    filename: String,
+}
+#[cfg(feature = "download-libtorch")]
+#[derive(serde::Deserialize, Debug)]
+struct PyPiPackage {
+    urls: Vec<PyPiPackageUrl>
+}
+#[cfg(feature = "download-libtorch")]
+fn get_pypi_wheel_url_for_aarch64_macosx() -> anyhow::Result<String> {
+    let pypi_url = format!("https://pypi.org/pypi/torch/{TORCH_VERSION}/json");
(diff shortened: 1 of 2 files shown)


torch-sys/build.rs
```diff
@@ -118,19 +163,11 @@ fn prepare_libtorch_dir() -> PathBuf {
...
-                            You can install torch manually following the indications from https://github.com/LaurentMazare/tch-rs/issues/629
-                            pip3 install torch=={TORCH_VERSION}
-
-                            Then update the following environment variables:
-                            export LIBTORCH=$(python3 -c 'import torch; from pathlib import Path; print(Path(torch.__file__).parent)')
-                            export DYLD_LIBRARY_PATH=${{LIBTORCH}}/lib
-                            ");
-                        }
+                    if env::var("CARGO_CFG_TARGET_ARCH") == Ok(String::from("aarch64")) {
+                        get_pypi_wheel_url_for_aarch64_macosx().expect("Failed to python wheel url from pypi")
```
LaurentMazare, line 167: Could we put back the original error message in the expect here, this way users might have some workable failover to try and would know which github issue to look at.

LaurentMazare (approved):

antimora:
@LaurentMazare Looking for this fix to be released soon. When is your next scheduled release?

LaurentMazare:
Ah yeah I haven't done a released in a while so I just pushed a new one out (0.12.0). Note that it contains some breaking changes as per the [changelog](https://github.com/LaurentMazare/tch-rs/blob/main/CHANGELOG.md).

PR #666, opened 2023-04-01
Change Linear::forward() to match pytorch

adeschamps:
The current implementation of `Linear::forward()` is valid and correct; however it produces outputs
that are similar but not numerically identical to `torch.nn.Linear` in PyTorch. This commit replaces
the implementation with `Tensor::linear()`, which matches the PyTorch outputs exactly.

A Python script is included which generates a random linear layer and input tensor, then writes the
tensors and their expected output to disk. A Rust unit test then confirms that `tch-rs` produces the
same output. The test is marked `#[ignore]`, because the tensors themselves are not committed (they
would be about 3M, which feels a little large; if it's fine then they can be committed).

### DIFF
--- a/src/nn/linear.rs
+++ b/src/nn/linear.rs
@@ -46,10 +46,27 @@ pub fn linear<'a, T: Borrow<super::Path<'a>>>(
 
 impl super::module::Module for Linear {
     fn forward(&self, xs: &Tensor) -> Tensor {
-        if let Some(bias) = &self.bs {
-            xs.matmul(&self.ws.tr()) + bias
-        } else {
-            xs.matmul(&self.ws.tr())
-        }
+        xs.linear(&self.ws, self.bs.as_ref())
     }
+}
+
+#[test]
+/// To run this test, generate test tensors:
+///
+/// ```shell
+/// cd tests/linear
+/// python gen-test-data.py
+/// ```
+#[ignore]
+fn matches_pytorch() {
+    use crate::nn::Module;
+
+    let input = Tensor::read_npy("tests/linear/in.npy").unwrap();
+    let expected_output = Tensor::read_npy("tests/linear/out.npy").unwrap();
+    let ws = Tensor::read_npy("tests/linear/ws.npy").unwrap();
+    let bs = Tensor::read_npy("tests/linear/bs.npy").unwrap();
+
+    let linear = Linear { ws, bs: Some(bs) };
+    let output = linear.forward(&input);
+    assert!(output.allclose(&expected_output, 1e-5, 1e-8, false));
 }
--- a/tests/linear/gen-test-data.py
+++ b/tests/linear/gen-test-data.py
@@ -1,0 +1,12 @@
+import numpy as np
+import torch
+
+SIZE=512
+
+linear = torch.nn.Linear(SIZE, SIZE)
+input = torch.rand(SIZE, SIZE)
+
+np.save('ws.npy', linear.weight.data)
+np.save('bs.npy', linear.bias.data)
+np.save('in.npy', input.data)
+np.save('out.npy', linear(input).data)


src/nn/linear.rs
```diff
@@ -46,10 +46,27 @@ pub fn linear<'a, T: Borrow<super::Path<'a>>>(
...
+    use crate::nn::Module;
+
+    let input = Tensor::read_npy("tests/linear/in.npy").unwrap();
+    let expected_output = Tensor::read_npy("tests/linear/out.npy").unwrap();
+    let ws = Tensor::read_npy("tests/linear/ws.npy").unwrap();
+    let bs = Tensor::read_npy("tests/linear/bs.npy").unwrap();
+
+    let linear = Linear { ws, bs: Some(bs) };
+    let output = linear.forward(&input);
+    assert!(output.allclose(&expected_output, 1e-5, 1e-8, false));
```
LaurentMazare, line 71: Why aren't these exactly equal? Using allclose seems a bit fidly as it could have worked for the old implementation.

src/nn/linear.rs
```diff
@@ -46,10 +46,27 @@ pub fn linear<'a, T: Borrow<super::Path<'a>>>(
...
 }
+
+#[test]
+/// To run this test, generate test tensors:
+///
+/// ```shell
+/// cd tests/linear
+/// python gen-test-data.py
+/// ```
+#[ignore]
```
LaurentMazare, line 60: Is this ignore so that the test is not run by default? If that's the case maybe we could check in the repo the npy files so that it can always be run (in which case the size should probably be reduced in order for the resulting files to only be a couple KB).

src/nn/linear.rs
adeschamps: Good point; this was mostly out of habit. Also, I was aware of `Tensor::eq_tensor(_) -> Tensor`, but I had forgotten about `Tensor::equal(_) -> bool`.

src/nn/linear.rs
adeschamps: That's correct. I've struggled to find tensors smaller than 512x512 that would consistently reproduce the issue. Perhaps libtorch takes different code paths depending on the size of the tensors? I'm not sure, but that wouldn't surprise me.

With 512x512 tensors, I have this test passing, which shows that the current `matmul()` implementation is close to the PyTorch output, but not equal. Still, this requires about 1MB of test data.
```rust
    assert!(original_output.allclose(&expected_output, 1e-2, 1e-3, false));
    assert!(!original_output.equal(&expected_output));
    assert!(output.equal(&expected_output));
```

adeschamps: With some further trial and error / bisecting, I managed to get it down to a 257x4 linear layer and a 2x257 input tensor (which seems to support my guess that libtorch uses different code paths depending on tensor sizes). This adds up to 24k of test data, so I've gone ahead and committed it, and removed the `#[ignore]`.
14c06c2

LaurentMazare (approved):

LaurentMazare:
Hum actually the `matches_pytorch` test failed in CI, maybe the result are inconistent between cpu and gpu (if you're using the gpu version to develop), or maybe even across cpus. We should probably switch this back to a non strict comparison I'm afraid.
CI failure: https://github.com/LaurentMazare/tch-rs/actions/runs/4599620341/jobs/8125212344
```
    Finished test [unoptimized + debuginfo] target(s) in 2m 07s
     Running unittests src/lib.rs (target/debug/deps/tch-a7fa23461eb8c594)

running 5 tests
test tensor::npy::tests::parse ... ok
test tensor::safetensors::tests::parse ... ok
test wrappers::jit::tests::ivalue_round_trip ... ok
test wrappers::scalar::tests::scalar ... ok
test nn::linear::matches_pytorch ... FAILED

failures:

---- nn::linear::matches_pytorch stdout ----
thread 'nn::linear::matches_pytorch' panicked at 'assertion failed: output.equal(&expected_output)', src/nn/linear.rs:71:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    nn::linear::matches_pytorch

test result: FAILED. 4 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

adeschamps:
Interesting.... I was developing on CPU, but I'm using libtorch 1.13.1 which isn't the latest. I'll check a few variations.

joshgev:
I just tried it using the libtorch that `build.rs` pulls in and it succeeds.

PR #667, opened 2023-04-03
Rework linear test to not expect exact equality.

adeschamps:
Proposed follow up to the CI failure introduced by https://github.com/LaurentMazare/tch-rs/pull/666

The test using `Tensor::equal()` worked on _my_ machine, but failed in CI. This version of the test
asserts that the implementation using `linear()` produces output that is `allclose()` to the
expected output, and also asserts that the distance to the expected output is smaller for the
`linear()` implementation than for the original `matmul()` implementation.

### DIFF
--- a/src/nn/linear.rs
+++ b/src/nn/linear.rs
@@ -65,8 +65,10 @@ fn matches_pytorch() {
     let linear = Linear { ws, bs };
     let output = linear.forward(&input);
 
-    // The `matmul()` implementation close, but not as close as calling `linear()`
-    assert!(original_output.allclose(&expected_output, 1e-2, 1e-3, false));
-    assert!(!original_output.equal(&expected_output));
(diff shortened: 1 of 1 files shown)


adeschamps:
I updated this PR to use `<=` instead of `<`, which should make the tests pass. Your guess about differences between CPU architectures seems reasonable, given that the version with `<` passed on my machine but not in CI.

I find this a little unsatisfying, because now the test doesn't really prove that `linear()` gives closer results to LibTorch than `matmul()` did, just that it's not any _further_.... although I swear that the difference was significant on my machine! (particularly in the context of a larger model)

LaurentMazare:
Got it, thanks for testing. Even if it's not ideal, the new test doesn't seem too bad especially if this is closer to the actual python implementation.

LaurentMazare (approved):

PR #693, opened 2023-04-29
Fix number of element calculation for `Vec<T>` conversion

guillaume-be:
Hello,

Recent changes for the conversion of tensors to vectors are leading to issues for the tensor to ndarray conversion.
The latter relies on an intermediate flat `Vec` representation (regardless of the original tensor dimension) stored alongside the size of the tensor to create the ndarray. However the updated method for converting a `Tensor` to a `Vec<T>` is only correct if the input tensor is also of dimension 1.

This small PR ensures converting a `Tensor` to a `Vec<T>` always leads to the correct number of elements. Converting a higher dimensionality Tensor to a `Vec<T>` will lead to flatten view of the same (as was the case previously).

I have added a small unit test to ensure support of the conversion of `Tensor` to `ndarray` for dimensions 1, 2 and 3

### DIFF
--- a/src/tensor/convert.rs
+++ b/src/tensor/convert.rs
@@ -7,7 +7,7 @@ use std::convert::{TryFrom, TryInto};
 impl<T: Element + Copy> TryFrom<&Tensor> for Vec<T> {
     type Error = TchError;
     fn try_from(tensor: &Tensor) -> Result<Self, Self::Error> {
-        let num_elem = tensor.numel() as usize;
+        let num_elem = tensor.numel();
         let mut vec = vec![T::ZERO; num_elem];
         tensor.f_to_kind(T::KIND)?.f_copy_data(&mut vec, num_elem)?;
         Ok(vec)
--- a/src/tensor/convert.rs
+++ b/src/tensor/convert.rs
@@ -7,10 +7,17 @@ use std::convert::{TryFrom, TryInto};
 impl<T: Element + Copy> TryFrom<&Tensor> for Vec<T> {
     type Error = TchError;
     fn try_from(tensor: &Tensor) -> Result<Self, Self::Error> {
-        let num_elem = tensor.numel();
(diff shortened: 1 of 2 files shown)


LaurentMazare:
Thanks for catching this, and sorry for the breakage. I feel a bit that the old behavior which implicitely flatten the array is error prone hence introducing the new semantics and failing when the number of dimensions is different from expected. This part of the change was a conscious decision (as mentioned in the changelog) but I didn't notice that the ndarray conversion would suffer from this. If the main concern is ndarray maybe we could just fix the conversion there instead?

guillaume-be:
Makes sense - I will submit an amended version shortly

LaurentMazare:
Actually I feel a bit that this would have the same drawback as the old conversion to vec of implicitly flattening (and not be coherent with the vec implementation). Maybe instead the new behavior could be preserved for ndarray too and only the error message would be tweaked to mention something like "if you want to flatten the tensor before converting it, us `tensor.reshape(-1)`" or something like this. Fwiw the major version number of the crate has been changed so that a breaking change is not an issue.

guillaume-be:
I feel the `ndarray` conversion is different than the `Vec<T>` conversion. While I agree it is surprising to have a multi-dimensional tensor flattened automatically to a 1-dimensional `Vec<T>`, `ndarray::ArrayD` are, like tensors, multi-dimensional data representations.

The conversion of a `Tensor` to a `ndarray::ArrayD` creates an intermediate flattened `Vec<T>` element because it is the expected data storage for the array creation, but the output from the conversion remains a multi-dimensional data container (it takes the shape of the original tensor as an input), allowing converting N-dimensional tensors to N-dimensional arrays (similarly to how Pytorch allows creating a N-dimensional numpy array from a tensor).

LaurentMazare:
Ah I didn't realise that the resulting ndarray would have the proper shape, this seems all good then, thanks!

LaurentMazare (approved):

PR #694, opened 2023-04-29
Reduce the number of calls through the ffi.

LaurentMazare:


guillaume-be:
Hello @LaurentMazare 

Thank you for further improving this conversion method. I wanted to take the chance to share some challenges currently faced when performing a high number of these conversions as part of integrating ONNX runtime models with pre/post-processing using tch Tensors.

I am observing a ~10x penalty hit going from `tch` implementations to `onnx` models, and I suspect the conversion is one of the reasons for it. I am detailing the challenges faced at https://github.com/pykeio/ort/issues/46, your input would be greatly appreciated

Thank you!

PR #723, opened 2023-05-24
Python extension example requires pytorch be imported - add to pymodule

jmalicki:


jmalicki (commented):
By default, trying the simple python
extension without importing `torch` will SEGFAULT.

Since this code requires `torch` to be already initialized to work properly, it seems to make sense to have the `pymodule` import torch itself, to guarantee initialization.

LaurentMazare (approved):

LaurentMazare:
Makes sense indeed, thanks for the PR.

jmalicki:
@LaurentMazare thanks!

Just to clarify for your understanding, I modified it to create a tensor on the rust size then export it without ever importing torch on python, thus torch was uninitialized.

Sorry for the confusion... thanks for releasing this code only a few weeks ago, I'm already building very nontrivial rust extensions on top of it (async dataloaders where the python ones weren't able to max out GPUs).

The part of the code that segfaulted was `pyobject_wrap` , where it tried to look up the python type but it was a NULL pointer, as the initialization code  wasn't yet run, since pytorch wasn't yet imported.

Sorry my initial message didn't properly explain.. no action is requested, the outcome is good, I just want to explain the scenario and bug properly for future posterity, and education around this.

PR #733, opened 2023-05-30
add correction and update test for onehot()

GaryBoone:
I found that my code failed when run on a GPU because the tensor::onehot() fn was hardcoded to use Device::Cpu. 

This PR changes onehot() to use the same device as the tensor it is applied to. The onehot test is modifies to confirm the new behavior.

 In the test, the device is set to Mps to confirm that it is not defaulting to CPU. This test was run on a non-mps device, confirming that the test does not require Mps to run.

### DIFF
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -227,7 +227,7 @@ impl Tensor {
     /// [N1, ..., Nk, labels]. The returned tensor uses float values.
     /// Elements of the input vector are expected to be between 0 and labels-1.
     pub fn onehot(&self, labels: i64) -> Tensor {
-        Tensor::zeros([self.size(), vec![labels]].concat(), crate::wrappers::kind::FLOAT_CPU)
+        Tensor::zeros([self.size(), vec![labels]].concat(), (self.kind(), self.device()))
             .scatter_value_(-1, &self.unsqueeze(-1).to_kind(Kind::Int64), 1.0)
     }
 
--- a/src/tensor/mod.rs
+++ b/src/tensor/mod.rs
@@ -227,7 +227,7 @@ impl Tensor {
     /// [N1, ..., Nk, labels]. The returned tensor uses float values.
     /// Elements of the input vector are expected to be between 0 and labels-1.
     pub fn onehot(&self, labels: i64) -> Tensor {
-        Tensor::zeros([self.size(), vec![labels]].concat(), (self.kind(), self.device()))
+        Tensor::zeros([self.size(), vec![labels]].concat(), (Kind::Float, self.device()))
             .scatter_value_(-1, &self.unsqueeze(-1).to_kind(Kind::Int64), 1.0)
     }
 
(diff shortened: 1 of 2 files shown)


src/tensor/mod.rs
```diff
@@ -227,7 +227,7 @@ impl Tensor {
     /// [N1, ..., Nk, labels]. The returned tensor uses float values.
     /// Elements of the input vector are expected to be between 0 and labels-1.
     pub fn onehot(&self, labels: i64) -> Tensor {
-        Tensor::zeros([self.size(), vec![labels]].concat(), crate::wrappers::kind::FLOAT_CPU)
+        Tensor::zeros([self.size(), vec![labels]].concat(), (self.kind(), self.device()))
```
LaurentMazare, line 230: I think this function is supposed to return a float rather than the kind of the input tensor (see the documentation).

tests/tensor_tests.rs
```diff
@@ -132,13 +132,25 @@ fn cat_and_stack() {
 
 #[test]
 fn onehot() {
```
LaurentMazare, line 134: Could we keep the original test as it was already helpful, and maybe add a new test that is protected by `has_mps` (as I don't know of any guarantees when it comes to using the mps device when `has_mps` returns false)?

src/tensor/mod.rs
GaryBoone: I found documentation [here](https://pytorch.org/docs/stable/generated/torch.nn.functional.one_hot.html) that says it should return a tensor of longs. Interestingly, `onehot()` is not listed in the `nn::functional` documentation [here](https://pytorch.org/docs/stable/nn.functional.html), nor found by searching from within the site. 

In this case, however, the `onehot()` function is defined on tensor, so I would expect the resulting onehot tensor to be the same type as the tensor it is applied to.

tests/tensor_tests.rs
GaryBoone: I dropped the `Mps` test as I don't have an Mps machine atm to confirm the test. I modified the test to show that it at least applies to both int and float tensors.

src/tensor/mod.rs
LaurentMazare: I would rather keep the old behavior for now as it's a guarantee of the current api and just change the device as for this one it seems pretty clear that it should be preserved (I would have to think more about it but I don't see a scenario where one would want to keep an integer type for the onehot encoded - hence taking more time to think about this/gather potential users feedback).

tests/tensor_tests.rs
GaryBoone: done.

src/tensor/mod.rs
GaryBoone: Ok, done. Restored the Float kind. Now the only change to the test is to confirm that the self and output tensors are on the same device.

LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #734, opened 2023-06-01
Added mps related functions for device

msminhas93:
The device.rs has Mps device, however it doesn't have utility functions associated with it. Added is_available, mps_if_available, is_mps functions similar to Cuda.


msminhas93:
@LaurentMazare any thoughts?

PR #738, opened 2023-06-06
Update README.md

Mottl:
Adds a clause about installation on Mac

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,12 @@ This version is then linked against.
 - Get `libtorch` from the
 [PyTorch website download section](https://pytorch.org/get-started/locally/) and extract
 the content of the zip file.
-- For Linux users, add the following to your `.bashrc` or equivalent, where `/path/to/libtorch`
+- For Linux and Mac users, add the following to your `.bashrc` or equivalent, where `/path/to/libtorch`
 is the path to the directory that was created when unzipping the file.
 ```bash
 export LIBTORCH=/path/to/libtorch
-export LD_LIBRARY_PATH=${LIBTORCH}/lib:$LD_LIBRARY_PATH
+export LD_LIBRARY_PATH=${LIBTORCH}/lib:$LD_LIBRARY_PATH # for Linux
+export DYLD_LIBRARY_PATH=${LIBTORCH}/lib:$DYLD_LIBRARY_PATH # for Mac
 ```
 The header files location can also be specified separately from the shared library via
 the following:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,11 +46,12 @@ This version is then linked against.
...
 [PyTorch website download section](https://pytorch.org/get-started/locally/) and extract
 the content of the zip file.
-- For Linux users, add the following to your `.bashrc` or equivalent, where `/path/to/libtorch`
+- For Linux and Mac users, add the following to your `.bashrc` or equivalent, where `/path/to/libtorch`
 is the path to the directory that was created when unzipping the file.
 ```bash
 export LIBTORCH=/path/to/libtorch
-export LD_LIBRARY_PATH=${LIBTORCH}/lib:$LD_LIBRARY_PATH
+export LD_LIBRARY_PATH=${LIBTORCH}/lib:$LD_LIBRARY_PATH # for Linux
+export DYLD_LIBRARY_PATH=${LIBTORCH}/lib:$DYLD_LIBRARY_PATH # for Mac
```
LaurentMazare, line 54: Do you find this to be actually necessary with the current github tip version of tch? We've recently made some changes setting up the rpath on linking so that at least on linux there is no need to set `LD_LIBRARY_PATH` but I haven't tried it out on a mac.

Mottl: It doesn't work on mac, at least out of the box. I need to set `DYLD_LIBRARY_PATH`

LaurentMazare: That's sad, I've left some mention to the library path this in the FAQ part of the readme for this case but it certainly seems like something that should be fixed and I don't have access to a mac to try it out (I'm more and more thinking about getting one just because of the amount of issues that get reported on macOS).

Just to be sure, you're using the latest github version right and building the examples right?

Could you check with otool whether the generated binaries have the proper rpath being set? More details [here](https://stackoverflow.com/questions/12521802/print-rpath-of-an-executable-on-macos).

Mottl: Thank you, Laurent!

cargo can build it successfully but I get the error while running without `DYLD_LIBRARY_PATH`:
```
dyld[8550]: Library not loaded: '@rpath/libtorch_cpu.dylib'
  Referenced from: 'mypath/target/release/myexecutable'
  Reason: tried: '/usr/local/lib/libtorch_cpu.dylib' (no such file), '/usr/lib/libtorch_cpu.dylib' (no such file)
Abort trap: 6
```

`otool -l` shows this:
```
Load command 15
          cmd LC_LOAD_DYLIB
      cmdsize 48
         name @rpath/libtorch.dylib (offset 24)
   time stamp 2 Thu Jan  1 07:00:02 1970
      current version 0.0.0
compatibility version 0.0.0
```

Issue #461, opened 2022-03-10
I don't know much about this question, but I also met this question when I want to use the same tensor in multiple threads.  After a lot of trying, I used a TensorWrapper struct to wrap the tensor and then impl send and sync for the struct. It works at last.


neveranever98:
I don't know much about this question, but I also met this question when I want to use the same tensor in multiple threads.  After a lot of trying, I used a TensorWrapper struct to wrap the tensor and then impl send and sync for the struct. It works at last.
`Struct TensorWrapper{tensor:Arc<Tensor>}`
`unsafe impl Send for TensorWrapper{}`
`unsafe impl Sync for TensorWrapper{}`

Hope this can solve your problem.
The tensor wrapped in TensorWrapper is just used to train a model thus I won't make any changes to this tensor, I'm also curious about the danger this could pose. I am also a beginner of rust so I don't know whether !sync of tensor is because of the possible danger of sync Tensor or because the underlying design of Rust makes sync isn't unnecessary for Tensor.

_Originally posted by @neveranever98 in https://github.com/LaurentMazare/tch-rs/issues/460#issuecomment-1064077067_

any35:
I don't think Tensor is thread safe in pytorch. 

In general, all the objects in pytorch are thread safe to read. But are not to write into.

Issue #460, opened 2022-03-03
How to inference in multi threads?


liebkne:
Sorry for dumb Rust newbie question. 

I understand the model can be executed using multiple threads by setting `tch::set_num_threads`. But what I'm trying to ask is can the model be used in multiple threads doing individual inference? 

I tried to put the model in `RwLock<SomeModel>` in app data of a actix server, but got errors like `help: within SomeModel, the trait Sync is not implemented for *mut torch_sys::C_tensor`. I read some materials like [1](https://discuss.pytorch.org/t/is-inference-thread-safe/88583), [2](https://discuss.pytorch.org/t/is-pytorch-supposed-to-be-thread-safe/36540/2) and it seems it should be possible in Python or C++. Am I right? If yes, can the Rust binding work with this situation?

Thanks.

neveranever98:
I don't know much about this question, but I also met this question when I want to use the same tensor in multiple threads.  After a lot of trying, I used a TensorWrapper struct to wrap the tensor and then impl send and sync for the struct. It works at last.
`Struct TensorWrapper{tensor:Arc<Tensor>}
unsafe impl Send for TensorWrapper{}
unsafe impl Sync for TensorWrapper{}
`
Hope this can solve your problem.
The tensor wrapped in TensorWrapper is just used to train a model thus I won't make any changes to this tensor, I'm also curious about the danger this could pose. I am also a beginner of rust so I don't know whether !sync of tensor is because of the possible danger of sync Tensor or because the underlying design of Rust makes sync isn't unnecessary for Tensor.

any35:
For now, I don't think the torch is ready for multi-thread service like web service.

For your scenario, run your model in a single thread (or a model per thread),  
Then try to retrieve your own data or parameter inter thread NOT tensor.

There are many ways to pass your own data, e.g., mspc.


These should work.

PR #751, opened 2023-07-07
Implemented sync

GSPuniani:
When using tch-rs inside a threaded context, `Sync` needs to be implemented. I have added the necessary trait bounds here to accomplish this.


LaurentMazare:
There has already been some discussions around this in the past and the consensus seems to be that tensors are not thread-safe enough to be allowed to be `Sync` (even `Send` might be a bit dodgy at the moment). see e.g. #460 and #461.

GSPuniani:
I see. That makes sense, thank you.

PR #759, opened 2023-07-16
Enable the function linalg_vector_norm

ramon-garcia:
This pull request is to enable the function linalg_vector_norm, disable in the original code.

I need it to compute the L2 norm of matrices and vectors.

It looks like it was disabled because of compile errors, caused by the removal of the parameter ord in the code generator, because it had a default value. After removing the default_value attribute in Declarations-v2.0.0.yaml, the code is correctly generated.

Of course, there are better solutions, but I wasn't able to make changes in the Ocaml code in gen/gen.ml.

Thank you for your time.

### DIFF
--- a/gen/gen.ml
+++ b/gen/gen.ml
@@ -37,7 +37,6 @@ let excluded_functions =
     ; "stride"
     ; "_assert_async"
     ; "gradient"
-    ; "linalg_vector_norm"
     ; "linalg_vector_norm_out"
     ; "linalg_matrix_norm"
     ; "linalg_matrix_norm_out"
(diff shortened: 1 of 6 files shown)


ramon-garcia:
ping

LaurentMazare:
Sorry for the delayed answer, making changes to the `Definitions.yaml` file would be a bit tricky to maintain when upgrading to new torch versions. If your goal is to compute the l2 norm, maybe you can use `mse_loss` or compute it directly via something like `x.pow_tensor_scalar(2).sum(Kind::float).sqrt()`?

ramon-garcia:
Thanks, I overlooked the fact that Definitions.yaml is autogenerated. I am going to review the patch to change the code generator instead.

LaurentMazare:
I would still encourage you to try the manual computations from my previous post if you can, it's certainly easier to get working and should cover all needs of norm computations that I can think of.

PR #775, opened 2023-07-26
Make indexing with empty ranges consistent

80Ltrumpet:
This brings the `IndexOp` functionality more in line with the analogous operation in Python, rather than panicking when the computed length of the range is negative.

```py
>>> import torch
>>> x = torch.randn([2, 3, 4])
>>> x[:, 1:1, :]
tensor([], size=(2, 0, 4))
>>> x[:, 1:0, :]
tensor([], size=(2, 0, 4))
```

Prior to this change, the second `let` below would cause a panic.

```rust
// Assume `x` is a `Tensor` like the one in the above Python example.
let good = x.i((.., 1..1, ..));
let oops = x.i((.., 1..0, ..));
```

Of course, this only causes a problem when the bounds of the range are computed at run time. :slightly_smiling_face:

### Is this a breaking change?

Arguably, no—unless someone can make the case that existing code relies on panics. Logic that potentially produces a zero-length empty range should also appropriately handle the logically empty `Tensor` resulting from indexing with it or any other empty range.

### DIFF
--- a/src/tensor/index.rs
+++ b/src/tensor/index.rs
@@ -303,32 +303,28 @@ impl Tensor {
                     curr_tensor.select(curr_idx, *index),
                     curr_idx, // not advanced because select() squeezes dimension
                 ),
-                Narrow(Unbounded, Unbounded) => (curr_tensor, curr_idx + 1),
-                Narrow(Included(start), Unbounded) => {
-                    let dim_len = curr_tensor.size()[curr_idx as usize];
-                    (curr_tensor.narrow(curr_idx, *start, dim_len - *start), curr_idx + 1)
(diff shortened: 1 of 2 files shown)


LaurentMazare (approved):

LaurentMazare:
Looks good, thanks for the PR! There is a CI failure because of clippy not liking reversed ranges like this, could you have a look at it before we merge?

80Ltrumpet:
Oops! I made sure to run `clippy` on the tests this time. Thank you for taking a look!

LaurentMazare:
Merged, thanks for the PR!

PR #793, opened 2023-08-28
Increase version of C++ used on torch-sys

dayo05:
In manually-compiled pytorch seems like requires C++17 capable features.

This is the error which I got with manual complicated code with M1 macs
[Log](https://pastebin.com/MNvcpB3a)

This PR only increases the version used on torch-sys building task.


LaurentMazare:
Somehow this ended up being changed as it was required by a recent libtorch update.

PR #801, opened 2023-09-24
Add support for `Dict[str, Any]`

VivekPanyam:
The current `GenericDict` impl either converts to `Dict[str, Tensor]` or `Dict[Any, Any]`. This PR allows it to convert to `Dict[str, Any]` as well (which is much easier to use within a model than `Dict[Any, Any]`).


VivekPanyam:
@LaurentMazare can you take a look at this please?

LaurentMazare (approved):

LaurentMazare:
Looks good, merged (and sorry for the delay).

PR #811, opened 2023-10-23
Add imagenet 21k labels

SmarakNayak:
Adds `CLASS_COUNT_21K` `CLASSES_21k` constants and `top_21k()` function to the `imagenet` library.

Labels were found [here](https://github.com/google-research/big_transfer/issues/7)

Probably a prettier way to do this.


LaurentMazare:
This adds a lot of code to the repo so I'm a bit afraid of the additional size + compile time etc. Would you mind making a separate repo and crate for this? Hopefully that should make for a trivial crate that wouldn't need a lot of maintenance and that everyone can re-use (even people that don't use candle).

SmarakNayak:
Yeah that makes sense, will let you know when the repo/crate is up

PR #842, opened 2024-01-31
Update PyO3 Bindings

CallMeMSL:
The PyO3 version used in this repo is not compatible with the latest. This pr updates the dependency to PyO3 and removes a now unnecessary import. I also added 2 little conversion tests, which are not executed by the ci (this would require additional ci config for the Python interpreter).


CallMeMSL:
It might also be a good idea to keep the version of `pyo3-tch` in sync with `pyo3`, not `tch`

Issue #851, opened 2024-02-28
Build fails unless I downgrade `google-glog` to 0.6.0.


iago-lito:
With this simple project:

- `Cargo.toml`
```toml
[package]
name = "tch_test"
version = "0.1.0"
edition = "2021"

[dependencies]
tch = { version = "0.15.0", features = ["download-libtorch"] } 
```

- `src/main.rs`
```rust
fn main() {}
```

After a clean `rustup update` and `sudo pacman -Syu` on archlinux, I get the following build failure:
```sh
$ cargo check
```

[`cargo_check.log`](https://github.com/LaurentMazare/tch-rs/files/14434876/cargo_check.log)

The build suceeds if I downgrade my [google-glog](https://archlinux.org/packages/extra/x86_64/google-glog/) package from `0.7.0` to `0.6.0`.

I can read nothing [on the README page](https://github.com/LaurentMazare/tch-rs) that refers to this library. Maybe it should be specified that it is actually used/useful? And that a specific version of it should be available on the system?

iago-lito:
Interestingly, the build also succeeds if I uninstall [`python-pytorch-cuda`](https://archlinux.org/packages/extra/x86_64/python-pytorch-cuda/) from my system. I guess the `download-libtorch` feature is actually _ignored_ if there is a libtorch available on the system, is it?
Not sure what's the best then. Enable `download-libtorch` "forcing"? Or maybe bump the expected libtorch/glog versions?

LaurentMazare:
No clue why google-glog is used, most likely is a transitive dependency so you will have to inspect your `Cargo.lock` file to get a sense of which package actually requires it.
I would recommend against using `download-libtorch` and install libtorch manually instead. This automatic download is more of an historical mistake than anything else and should be avoided if you can.

Jark5455:
Google GLOG was recently updated and that update was pushed out to the arch repo. This update requires that when you include GLOGs headers you either have `GLOG_EXPORT` or `GLOG_NO_EXPORT` defined.

blmarket:
> I would recommend against using `download-libtorch` and install libtorch manually instead. This automatic download is more of an historical mistake than anything else and should be avoided if you can.

Manually installed libtorch still have this problem. At least with Arch Linux.

[...]

PR #852, opened 2024-03-04
Temporary Fix for GLOG

Jark5455:
Here is a temporary fix for #851 that appears to work when tested on my arch box. In the future would strongly recommend excluding GLOG in the bindings generator.


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #893, opened 2024-09-12
add method to set default kind of new variables in VarStore

jquesnelle:
This changes the behavior of `VarStore::set_kind` to change the default kind for all new variables created in the `VarStore`, similar to `torch.set_default_dtype` (but localized to the `VarStore`).

The need for this is that for larger models like Llama-7b first instantiating an empty model and then filling in the tensors from an mmap-ed safetenstors file (for example) won't work on a 24GB GPU since a 7B won't fit in `Kind::Float`. Rather, newly inited tensors need to start out as `Kind::Half` or `Kind::BFloat16`.

The default behavior of `VarStore` is unchanged -- existing code that doesn't call `set_kind` will have new tensors created at as `Kind::Float`.

### DIFF
--- a/src/nn/init.rs
+++ b/src/nn/init.rs
@@ -113,9 +113,7 @@ pub fn f_init(i: Init, dims: &[i64], device: Device, kind: Kind) -> Result<Tenso
                 Tensor::f_ones(dims, (kind, device)).map(|t| t * cst)
             }
         }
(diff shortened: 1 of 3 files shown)


LaurentMazare (approved):

LaurentMazare:
Thanks!

PR #894, opened 2024-09-14
npy: fix parsing of complex dtypes

japm48:
An exported `.npy` file with complex single float dtype fails to be loaded with `FileFormat("unrecognized descr c8")` (similarly with `c16` for complex double).
According to [the documentation](https://numpy.org/doc/2.1/reference/arrays.dtypes.html), `c{size}` are the only values that should be parsed as complex. I could not find any reference of `F` or `D` as introduced in #468.

### DIFF
--- a/src/tensor/npy.rs
+++ b/src/tensor/npy.rs
@@ -140,8 +140,8 @@ impl Header {
                     "b" | "i1" => Kind::Int8,
                     "B" | "u1" => Kind::Uint8,
                     "?" | "b1" => Kind::Bool,
-                    "F" | "F4" => Kind::ComplexFloat,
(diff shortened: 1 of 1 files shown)


LaurentMazare:
I feel that we should keep the support for `F` and `D` just in case as it's for reading only and is not used by any other dtype. I can find some references to it [here](https://github.com/numpy/numpy/blob/b307a8c89efc3c2878eb2820b98e8de58d68771c/numpy/lib/_type_check_impl.py#L575-L576) for example though it's hard to tell if it can end up in a `.npy` file.

japm48:
Thanks for the research. I updated the PR accordingly.

LaurentMazare (approved):

PR #899, opened 2024-10-10
Fix glog for pytorch 2.3.x branch.

de-vri-es:
Sadly I can't update to pytorch 2.4.0 yet, but glog is already broken (same problem as #852).

So I would be very happy if we can backport this to the 2.3.0 branch and have a release so I can get back to using a semver dependency <3

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tch"
-version = "0.16.0"
+version = "0.16.1"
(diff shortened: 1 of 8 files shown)


LaurentMazare:
Merged, I've published some minor update to the 0.16 release based on your branch so hopefully you can just use them now, e.g. https://crates.io/crates/tch/0.16.1

de-vri-es:
Awesome! Thanks :heart:

PR #908, opened 2024-11-15
rfc: support for torch-tensorrt

hietalajulius:
# RFC: Add Torch-TensorRT Support

## Summary
This PR adds support for NVIDIA's Torch-TensorRT in tch-rs, enabling GPU inference optimization through TensorRT while maintaining the PyTorch ergonomics.

## Motivation
Torch-TensorRT is NVIDIA's official PyTorch-TensorRT integration that can provide:

- Up to 5x faster inference compared to eager execution
- Automatic optimization of PyTorch models
- Seamless integration with existing PyTorch workflows
- Support for both dynamic and static shapes

### Build System Changes

Added support through:

1. New torch-tensorrt feature flag in Cargo.toml
2. Build script modifications to link TensorRT libraries
3. C++ preprocessor flag USE_TORCH_TENSORRT

### Important Prerequisites & Caveats

1. Python Environment Requirement
    * Torch-TensorRT must be installed via pip in your Python environment:
`pip install torch-tensorrt`
    * The build script will detect TensorRT through this installation

2. Rust Nightly Requirement
    * Requires Rust nightly toolchain due to the use of the `-as-needed` linker option
    * Add to your project:
        ```
        [toolchain]
        channel = "nightly"
        ```

    * Run with `RUSTFLAGS="-Zunstable-options"`

3. Runtime Environment
* `LD_LIBRARY_PATH` must include TensorRT library paths
* E.g.: `export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/path/to/python3.12/site-packages/torch_tensorrt/lib:/path/to/python3.12/site-packages/tensorrt_libs`

## Usage
Enable in your project's Cargo.toml e.g.:
```
[features]
torch-tensorrt = ["tch/torch-tensorrt"]
```

## Questions
* Overall is this within the scope of `tch-rs`?
* Only having support on nightly is not great, this can be avoided by using e.g. `LD_PRELOAD` (https://pytorch.org/TensorRT/user_guide/runtime.html) to overcome the `-as-needed` issue which is not amazing either
* Installation via python is convenient in dev, but an actual production env would probably want to install the libs without the python dependency
* Further tweaking of the linker flags probably needed to avoid setting `LD_LIBRARY_PATH`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ download-libtorch = ["torch-sys/download-libtorch"]
 python-extension = ["torch-sys/python-extension"]
 rl-python = ["cpython"]
 doc-only = ["torch-sys/doc-only"]
+torch-tensorrt = ["torch-sys/torch-tensorrt"]
 cuda-tests = []
 
 [package.metadata.docs.rs]
--- a/torch-sys/Cargo.toml
+++ b/torch-sys/Cargo.toml
@@ -27,6 +27,7 @@ zip = "0.6"
 download-libtorch = ["ureq", "serde", "serde_json"]
 doc-only = []
 python-extension = []
+torch-tensorrt = []
 
 [package.metadata.docs.rs]
 features = [ "doc-only" ]
--- a/torch-sys/build.rs
+++ b/torch-sys/build.rs
@@ -412,9 +412,88 @@ impl SystemInfo {
     }
 }
 
+const PYTHON_PRINT_TORCH_TENSORRT_PATH: &str = r#"
+try:
+    import torch_tensorrt
+    import os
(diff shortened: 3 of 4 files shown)


hietalajulius:
@LaurentMazare let me know if this is something that could be relevant for `tch-rs`, otherwise I could consider building a new crate that just depends on `tch-rs` :+1:

hietalajulius:
After experimenting with this more, the easiest way to use Torrh-TensorRT with tch-rs is with e.g. `LD_PRELOAD=/path/to/libtorchtrt_runtime.so` (https://pytorch.org/TensorRT/user_guide/runtime.html), given how hard linking seems to be with `--no-as-needed` in rust and that none of the symbols are needed compile time for code using tch-rs. Closing.

PR #913, opened 2024-11-23
Support Retrieving Named Attributes

ammrat13:
Previously, this library supported getting a list of named tensors from a module. However, modules may also have attributes to store additional metadata. This commit adds support for getting them.

As for implementation, this commit follows the same strategy as retrieving named parameters, and it adds two smoke tests to check functionality. Importantly, this commit was tested with LibTorch 2.5.0 instead of 2.5.1, and `LIBTORCH_BYPASS_VERSION_CHECK` has to be set.


LaurentMazare:
Thanks for the PR and sorry for the delayed answer, would you mind making a much smaller `.pt` test file? Current one is ~700KB which is a lot to add to this repo.

PR #924, opened 2025-01-09
Add jit compile api

tye-singwa:
Hey @LaurentMazare! I noticed that `torch::jit::compile` is missing. Adding it in this pr
Is there anything i can improve?
Thanks


tye-singwa:
also adding `create_by_tracing_is`, as saving traced modules with multiple return tensors is not supported

PR #933, opened 2025-02-25
add more DINOv2 models

zhoubin-me:
- Add base, large, giant models
- Allow feature extraction without classifier's head


LaurentMazare:
Thanks!

PR #934, opened 2025-02-25
Add `c10::Device` to IValue enum

0dlinwaabi:
## Summary 
- Add `c10::Device` to `tch::IValue` enum. 
- The `c10::Device` is one of variant in the torch `IValue` enum. https://github.com/pytorch/pytorch/blob/d990dada86a8ad94882b5c23e859b88c0c255bda/aten/src/ATen/core/ivalue.h#L923


LaurentMazare:
Thanks!

PR #935, opened 2025-02-26
Add tensor slice indexing

OrigamiDream:
This PR aims to provide a `Slice(start, end, step)`. The following illustrates the usage.

```rust
use tch::{Tensor, IndexOp, Slice};

let tensor = Tensor::from_slice(&[1, 2, 3, 4, 5, 6]).view((2, 3));
let result = tensor.i((.., Slice::new(None, None, 2)));
```

It shows exactly same behavior in PyTorch: `tensor[:, ::2]`.

I had considered several ways of using `Slice`, such as:
```rust
let _ = Slice(0, -1, 2); // all exactly `i64`
let _ = Slice(Some(0), None, 2); // uses `Option<i64>` for `start` and `end`
let _ = Slice::new(0, None, 2); // add `new` method and uses `impl Into<Option<i64>>`
```

I ended up using the third one, which feels more natural to me. But I hope to discuss about this design more.

### DIFF
--- a/src/tensor/index.rs
+++ b/src/tensor/index.rs
@@ -44,6 +44,20 @@
 //! assert_eq!(t.size(), &[2, 3, 1]);
 //! ```
 //!
+//! The `Slice::new(start, end, step)` index can be used to select elements over step size.
+//!
+//! ```ignore
+//! use crate::tch::{IndexOp, Slice, Tensor};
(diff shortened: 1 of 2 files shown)


OrigamiDream:
Since this PR is abandoned, I would like to close this.

PR #960, opened 2025-06-03
Support for instantiating tensors from `num_complex::Complex` values.

### LINKED ISSUE #959: How to create a `Tensor` of `Complex<f32>`?
The following code fails to compile:

```rust
let mut buf = vec![Complex::<f32>::default(); 1024];
let t = Tensor::from_slice(buf.as_slice());
```

with error:

```
the trait bound `Complex<f32>: Element` is not satisfied`
```

There is a `Kind::ComplexFloat` variant, but no `impl Element for Complex<f32>`. Is this an oversight?

metasim:
Note: I implemented this under a `complex` feature flag assuming this should be opt-in. However, `ndarray` already has a fixed dependency on `num-complex` so there may be little benefit to doing so.

Closes #959.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ image = { version = "0.24.5", optional = true }
 clap = { version = "4.2.4", features = ["derive"], optional = true }
 serde_json = { version = "1.0.96", optional = true }
 memmap2 = { version = "0.6.1", optional = true }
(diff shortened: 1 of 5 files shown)


Dherse:
Any update missing from GH? This would be really useful :)

metasim:
@LaurentMazare Is there anything I can do to improve this PR to make it acceptable?

PR #1014, opened 2026-05-21
torch-sys: restore dummy CUDA dependency on Windows to fix Cuda::is_available

agene0001:
## Summary

On Windows MSVC builds, `tch::Cuda::is_available()` returns `false` and `device_count()` returns `0` even when a CUDA-capable GPU + driver + CUDA-flavored libtorch are all present. The same setup works correctly on Linux. This PR restores the Windows-relevant half of the mechanism removed in 16e8f59 (\"Attempt at removing the cuda hack\", Oct 2024).

## Root cause

MSVC's linker treats `torch_cuda.lib` as an import library: it only retains imports whose symbols are statically referenced by the consuming objects. Currently nothing in the compiled C++ stubs (`torch_api*.cpp`) or the generated Rust bindings touches a CUDA symbol, so:

1. `/OPT:REF` strips the unused import,
2. the final `.exe` PE header never lists `torch_cuda.dll` as a dependency,
3. the OS loader skips it at startup,
4. `torch_cuda.dll`'s `DllMain` never runs,
5. `at::globalContext().hasCUDA()` stays `false`,
6. `tch::Cuda::is_available()` always returns `false`.

Linux/macOS dodge this because ELF/Mach-O record `DT_NEEDED` from `-ltorch_cuda` regardless of symbol use — the shared object still loads at process start.

## Fix

Restore the pre-16e8f59 stub, but scope it to Windows + CUDA so the post-16e8f59 simplification stands on Linux/macOS:

- **`torch-sys/libtch/dummy_cuda_dependency.cpp`** — single-call stub. The original `getCurrentCUDABlasHandle` / `warp_size` symbols have been renamed/inlined in later libtorch releases and no longer link; `at::cuda::CachingHostAllocator_emptyCache` has been stable since PyTorch 2.3 and is a no-op when CUDA hasn't been used.
- **`torch-sys/build.rs`** — compile the stub only when `use_cuda && target_os = \"windows\"`. Emits `cargo:rustc-cfg=use_cuda` for crate-local gating and `cargo:cuda=1` metadata for downstream consumers (read by `tch` via `DEP_TCH_CUDA` thanks to `links = \"tch\"`).
- **`torch-sys/src/lib.rs`** — declares `extern \"C\" { fn dummy_cuda_dependency(); }` under `cfg(use_cuda)`.
- **`build.rs`** (top-level) — propagates the metadata, emitting `cargo:rustc-cfg=use_cuda` when `DEP_TCH_CUDA` is set.
- **`src/tensor/mod.rs`** — adds `#[used] static FORCE_TORCH_CUDA_LINK` referencing `torch_sys::dummy_cuda_dependency`, gated on `cfg(all(target_os = \"windows\", use_cuda))`. The anchor lives in `tensor/mod.rs` because every downstream consumer references `Tensor`, so that translation unit is always pulled out of the rlib and `#[used]` reliably keeps the static across rlib boundaries.

## Verification

Windows 11 + RTX 4070 + NVIDIA driver 595.97 + libtorch 2.9.1+cu126:

```
$ cargo run --example cuda_probe          # custom probe binary
tch::Cuda::is_available()       = true     # ← was false on main
tch::Cuda::device_count()       = 1        # ← was 0
tch::Cuda::cudnn_is_available() = true     # ← was false
CUDA tensor: [1.0, 2.0, 3.0]
```

No Linux/macOS code path changes — those platforms still rely on runtime-linker semantics as on `main`.

## Test plan

- [x] Windows + CUDA + cu126 libtorch: `Cuda::is_available() == true`, CUDA tensor round-trips.
- [ ] Windows + CUDA + cu128/cu130/cu132 libtorch (would appreciate help testing other CUDA flavors)
- [ ] Windows CPU-only libtorch: stub not compiled, anchor not emitted, unchanged behavior
- [ ] Linux + CUDA libtorch: cfg path skipped, unchanged behavior
- [ ] macOS: cfg path skipped, unchanged behavior

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -10,4 +10,11 @@ fn main() {
         }
         _ => {}
     }
+    // Propagate the CUDA-presence cfg set by `torch-sys/build.rs` (via the
+    // `links = "tch"` metadata channel) so we can gate the link anchor that
+    // forces MSVC to keep the `torch_cuda.lib` import.
+    println!("cargo:rustc-check-cfg=cfg(use_cuda)");
+    if std::env::var_os("DEP_TCH_CUDA").is_some() {
+        println!("cargo:rustc-cfg=use_cuda");
+    }
 }
--- a/src/data.rs
+++ b/src/data.rs
@@ -16,6 +16,7 @@ pub struct Iter2 {
     total_size: i64,
     device: Device,
     return_smaller_last_batch: bool,
+    pin_memory: bool,
 }
 
 impl Iter2 {
@@ -47,6 +48,7 @@ impl Iter2 {
             total_size,
             device: Device::Cpu,
             return_smaller_last_batch: false,
+            pin_memory: false,
         })
     }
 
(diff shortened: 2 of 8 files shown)


agene0001:
@sbeckeriv @LaurentMazare  any news on whether this could be pulled?

PR #1023, opened 2026-07-10
Use prebuilt libtorch on macOS as well

malfet:
Unify the macOS download path with Linux/Windows by fetching the prebuilt libtorch-macos-arm64 zip from download.pytorch.org instead of scraping a PyPI wheel.
Moreover, PyPI path was broken anyway: it looked for cp311-none-macosx_11_0_arm64 suffix that no longer matches what is published on PyPI.


getong:
I try to update https://github.com/tracel-ai/burn , which use tch 0.22, while I update tch to 0.24, the prebuilt issue in macos. 
Which is related to https://github.com/tracel-ai/burn/pull/5280 . 
Please merge this pr, and make a new release, thanks.

PR #1025, opened 2026-08-01
fix macos download name

getong:


getong:
#1023 solves this issue, close this.

